PR #18, opened 2021-11-16
CLI options

jwinarske:
When debug build CLI Options not parsed by main.cc are passed directly to Dart VM
Command line option '-a' overrides asset path'

Signed-off-by: Joel Winarske <joel.winarske@toyotaconnected.com>


cmc5788 (approved):
🎉

PR #51, opened 2022-04-21
xdg_shell, multi-output, remove sprawl, cleanup

jwinarske:
xdg_shell
-replaces wl_shell for xdg_shell

Multiple outputs
-Adds support for saving state on multiple outputs
-Enables deriving the buffer scaling factor

Cleanup
-Remove unused methods/variables
-Remove unused headers
-Initialize uninitialize variables
-rename m_engine to m_flutter_engine in App class

Remove sprawl option
-Not required with xdg_shell

Update README.md
-Enabling flutter desktop causes App build issue.
 This has been the case for some time.
 Only use custom devices for debugging eLinux Flutter.

FPS feature cleanup
-Fix narrowing warnings
-Update README for usage

Remove pinch gesture references
-Dart layer gesture detection being used.
 Not all targets have this protocol

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,21 @@ Copy libflutter_engine.so to `/usr/local/lib` or use LD_LIBRARY_PATH to point do
     export LD_LIBRARY_PATH=`pwd`:$LD_LIBRARY_PATH
     homescreen
 
+## FPS
+Three environmental variables impact FPS reporting.
+
+### 1. FPS_OUTPUT_CONSOLE
+
+Setting `FPS_OUTPUT_CONSOLE` to "1" will output FPS count to console.
+
+### 2. FPS_OUTPUT_OVERLAY
+
+If `FPS_OUTPUT_CONSOLE=1` and `FPS_OUTPUT_OVERLAY=1` the screen overlay is enabled.
+
+### 3. FPS_OUTPUT_FREQUENCY
+
+Changing `FPS_OUTPUT_FREQUENCY` controls the update interval.  This parameter is optional.
+
 ## Debug
 
 cd to flutter app folder
--- a/README.md
+++ b/README.md
@@ -125,11 +125,7 @@ Changing `FPS_OUTPUT_FREQUENCY` controls the update interval.  This parameter is
 
 ## Debug
 
-cd to flutter app folder
-
-    flutter config --enable-linux-desktop
-    flutter create .
-    flutter attach --debug-port 41795 --host-vmservice-port 41795
+Setup custom devices to control ivi-homescreen via debugger. 
 
 # CMAKE dependency paths
 
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -107,8 +107,9 @@ App::App(const std::string& app_id,
   m_fps_counter = 0;
 
   const char* envstr_console;
+  char* pEnd;
   if ((envstr_console = getenv("FPS_OUTPUT_CONSOLE")) != nullptr) {
-    int val = atoi(envstr_console);
+    long val = strtol(envstr_console, &pEnd, 10);
 
     if (0 < val) {
       m_fps_output |= 0x01;
@@ -117,7 +118,7 @@ App::App(const std::string& app_id,
 
   const char* envstr_overlay;
   if ((envstr_overlay = getenv("FPS_OUTPUT_OVERLAY")) != nullptr) {
-    int val = atoi(envstr_overlay);
+    long val = strtol(envstr_overlay, &pEnd, 10);
 
     if (0 < val) {
       m_fps_output |= 0x02;
@@ -126,7 +127,7 @@ App::App(const std::string& app_id,
 
   const char* envstr_freq;
   if ((envstr_freq = getenv("FPS_OUTPUT_FREQUENCY")) != nullptr) {
-    int val = atoi(envstr_freq);
+    long val = strtol(envstr_freq, &pEnd, 10);
 
     if (0 < val) {
       m_fps_period = val;
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -67,17 +67,17 @@ App::App(const std::string& app_id,
   }
 
   for (size_t i = 0; i < kEngineInstanceCount; i++) {
-    m_engine[i] = std::make_shared<Engine>(this, i, m_command_line_args_c,
+    m_flutter_engine[i] = std::make_shared<Engine>(this, i, m_command_line_args_c,
                                            application_override_path);
-    m_engine[i]->Run(pthread_self());
+    m_flutter_engine[i]->Run(pthread_self());
 
-    if (!m_engine[i]->IsRunning()) {
+    if (!m_flutter_engine[i]->IsRunning()) {
       FML_LOG(ERROR) << "Failed to Run Engine";
       exit(-1);
     }
 
     // Set Flutter Window Size
-    auto result = m_engine[i]->SetWindowSize(m_egl_window[i]->GetHeight(),
+    auto result = m_flutter_engine[i]->SetWindowSize(m_egl_window[i]->GetHeight(),
                                              m_egl_window[i]->GetWidth());
     if (result != kSuccess) {
       FML_LOG(ERROR) << "Failed to set Flutter Engine Window Size";
@@ -87,13 +87,13 @@ App::App(const std::string& app_id,
   }
 
   // Enable pointer events
-  m_display->SetEngine(m_engine[0]);
+  m_display->SetEngine(m_flutter_engine[0]);
 
 #ifdef ENABLE_TEXTURE_TEST
-  m_texture_test->SetEngine(m_engine[0]);
+  m_texture_test->SetEngine(m_flutter_engine[0]);
 #endif
 #ifdef ENABLE_PLUGIN_TEXT_INPUT
-  m_text_input->SetEngine(m_engine[0]);
+  m_text_input->SetEngine(m_flutter_engine[0]);
   m_display->SetTextInput(m_text_input);
 #endif
 
@@ -151,7 +151,7 @@ int App::Loop() {
                         std::chrono::steady_clock::now().time_since_epoch())
                         .count();
 
-  for (auto& i : m_engine) {
+  for (auto& i : m_flutter_engine) {
     i->RunTask();
   }
 
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -89,6 +89,9 @@ App::App(const std::string& app_id,
   // Enable pointer events
   m_display->SetEngine(m_flutter_engine[0]);
 
+  // Buffer Scaling
+  m_display->SetSurface(m_egl_window[0]->GetNativeSurface());
+
 #ifdef ENABLE_TEXTURE_TEST
   m_texture_test->SetEngine(m_flutter_engine[0]);
 #endif
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -36,9 +36,10 @@ App::App(const std::string& app_id,
          const std::string& cursor_theme_name)
     : m_gl_resolver(std::make_shared<GlResolver>()),
       m_display(
-          std::make_shared<Display>(this, enable_cursor, cursor_theme_name)),
+          std::make_shared<Display>(enable_cursor, cursor_theme_name)),
       m_egl_window{std::make_shared<EglWindow>(0,
                                                m_display,
+                                               m_display->GetBaseSurface(),
                                                EglWindow::WINDOW_BG,
                                                app_id,
                                                fullscreen,
@@ -57,7 +58,7 @@ App::App(const std::string& app_id,
 
   FML_DLOG(INFO) << "+App::App";
 
-  m_display->AglShellDoBackground(m_egl_window[0]->GetNativeSurface());
+  m_display->AglShellDoBackground(m_egl_window[0]->GetNativeSurface(), 0);
 
   std::vector<const char*> m_command_line_args_c;
   m_command_line_args_c.reserve(command_line_args.size());
@@ -67,7 +68,8 @@ App::App(const std::string& app_id,
   }
 
   for (size_t i = 0; i < kEngineInstanceCount; i++) {
-    m_flutter_engine[i] = std::make_shared<Engine>(this, i, m_command_line_args_c,
+    m_flutter_engine[i] = std::make_shared<Engine>(this, i,
+                                                   m_command_line_args_c,
                                            application_override_path);
     m_flutter_engine[i]->Run(pthread_self());
 
@@ -75,22 +77,13 @@ App::App(const std::string& app_id,
       FML_LOG(ERROR) << "Failed to Run Engine";
       exit(-1);
     }
-
-    // Set Flutter Window Size
-    auto result = m_flutter_engine[i]->SetWindowSize(m_egl_window[i]->GetHeight(),
-                                             m_egl_window[i]->GetWidth());
-    if (result != kSuccess) {
-      FML_LOG(ERROR) << "Failed to set Flutter Engine Window Size";
-    }
+    m_egl_window[i]->SetEngine(m_flutter_engine[i]);
 
     FML_DLOG(INFO) << "(" << i << ") Engine running...";
   }
 
   // Enable pointer events
   m_display->SetEngine(m_flutter_engine[0]);
-
-  // Buffer Scaling
-  m_display->SetSurface(m_egl_window[0]->GetNativeSurface());
 
 #ifdef ENABLE_TEXTURE_TEST
   m_texture_test->SetEngine(m_flutter_engine[0]);
@@ -107,28 +100,27 @@ App::App(const std::string& app_id,
   m_fps_period = 1;
   m_fps_counter = 0;
 
-  const char* envstr_console;
-  char* pEnd;
-  if ((envstr_console = getenv("FPS_OUTPUT_CONSOLE")) != nullptr) {
-    long val = strtol(envstr_console, &pEnd, 10);
+  const char* env_string_console;
+  if ((env_string_console = getenv("FPS_OUTPUT_CONSOLE")) != nullptr) {
+    long val = strtol(env_string_console, nullptr, 10);
 
     if (0 < val) {
       m_fps_output |= 0x01;
     }
   }
 
-  const char* envstr_overlay;
-  if ((envstr_overlay = getenv("FPS_OUTPUT_OVERLAY")) != nullptr) {
-    long val = strtol(envstr_overlay, &pEnd, 10);
+  const char* env_string_overlay;
+  if ((env_string_overlay = getenv("FPS_OUTPUT_OVERLAY")) != nullptr) {
+    long val = strtol(env_string_overlay, nullptr, 10);
 
     if (0 < val) {
       m_fps_output |= 0x02;
     }
   }
 
-  const char* envstr_freq;
-  if ((envstr_freq = getenv("FPS_OUTPUT_FREQUENCY")) != nullptr) {
-    long val = strtol(envstr_freq, &pEnd, 10);
+  const char* env_string_freq;
+  if ((env_string_freq = getenv("FPS_OUTPUT_FREQUENCY")) != nullptr) {
+    long val = strtol(env_string_freq, nullptr, 10);
 
     if (0 < val) {
       m_fps_period = val;
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -31,7 +31,6 @@ App::App(const std::string& app_id,
          bool fullscreen,
          bool enable_cursor,
          bool debug_egl,
-         bool sprawl,
          uint32_t width,
          uint32_t height,
          const std::string& cursor_theme_name)
@@ -44,7 +43,6 @@ App::App(const std::string& app_id,
                                                app_id,
                                                fullscreen,
                                                debug_egl,
-                                               sprawl,
                                                width,
                                                height)}
 #ifdef ENABLE_TEXTURE_TEST
--- a/shell/display.cc
+++ b/shell/display.cc
@@ -41,7 +41,6 @@ Display::Display([[maybe_unused]] App* app,
       m_keymap(nullptr),
       m_xkb_state(nullptr),
       m_agl_shell(nullptr),
-      m_has_xrgb(false),
       m_enable_cursor(enable_cursor),
       m_cursor_theme_name(std::move(cursor_theme_name)) {
   FML_DLOG(INFO) << "+ Display()";
@@ -82,10 +81,6 @@ Display::Display([[maybe_unused]] App* app,
     FML_LOG(INFO) << "No agl_shell extension present";
   }
 
-  if (!m_has_xrgb) {
-    FML_LOG(INFO) << "WL_SHM_FORMAT_XRGB32 not available";
-  }
-
   FML_DLOG(INFO) << "- Display()";
 }
 
@@ -117,11 +112,12 @@ Display::~Display() {
   FML_DLOG(INFO) << "- ~Display()";
 }
 
-void Display::registry_handle_global(void* data,
-                                     [[maybe_unused]] struct wl_registry* registry,
-                                     [[maybe_unused]] uint32_t name,
-                                     [[maybe_unused]] const char* interface,
-                                     [[maybe_unused]] uint32_t version) {
+void Display::registry_handle_global(
+    void* data,
+    [[maybe_unused]] struct wl_registry* registry,
+    [[maybe_unused]] uint32_t name,
+    [[maybe_unused]] const char* interface,
+    [[maybe_unused]] uint32_t version) {
   auto* d = static_cast<Display*>(data);
 
   FML_DLOG(INFO) << "Wayland: " << interface;
@@ -131,7 +127,7 @@ void Display::registry_handle_global(void* data,
         wl_registry_bind(registry, name, &wl_compositor_interface, 1));
   }
 
-  if (strcmp(interface, wl_subcompositor_interface.name) == 0) {
+  else if (strcmp(interface, wl_subcompositor_interface.name) == 0) {
     d->m_subcompositor = static_cast<struct wl_subcompositor*>(
         wl_registry_bind(registry, name, &wl_subcompositor_interface, 1));
   }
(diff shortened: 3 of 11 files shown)


shell/egl_window.cc
```diff
@@ -52,18 +51,23 @@ EglWindow::EglWindow(size_t index,
...
+      xdg_wm_base_get_xdg_surface(m_display->GetXdgWmBase(), m_surface);
+  assert(m_xdg_surface);
+  FML_DLOG(INFO) << "EglWindow::m_xdg_surface = "
+                 << static_cast<void*>(m_xdg_surface);
+
+  xdg_surface_add_listener(m_xdg_surface, &xdg_surface_listener, this);
+  //  xdg_surface_add_listener(m_xdg_surface, &xdg_surface_listener,this);
+  m_xdg_toplevel = xdg_surface_get_toplevel(m_xdg_surface);
+  xdg_toplevel_set_title(m_xdg_toplevel, m_app_id.c_str());
+  //  wl_surface_commit(m_surface);
```
mv0: Better to just remove it entirely.

shell/display.cc
```diff
@@ -170,7 +170,7 @@ void Display::registry_handle_global(
 
   else if (strcmp(interface, wl_output_interface.name) == 0) {
     d->m_output = static_cast<struct wl_output*>(
-        wl_registry_bind(registry, name, &wl_output_interface, 1));
+        wl_registry_bind(registry, name, &wl_output_interface, 2));
```
mv0: For the versioning we usually do MIN(2, version).

shell/egl_window.cc
```diff
@@ -107,9 +107,9 @@ EglWindow::EglWindow(size_t index,
   m_fps_idx = 0;
   m_fps_counter = 0;
 
-  //  this->m_callback = wl_surface_frame(this->m_surface);
-  //  wl_callback_add_listener(this->m_callback, &frame_listener, this);
-  wl_surface_commit(this->m_surface);
+  m_callback = wl_surface_frame(this->m_surface);
+  wl_callback_add_listener(m_callback, &frame_listener, this);
+  wl_surface_commit(m_surface);
```
mv0: Probably this wouldn't need any comments from the previous patch. Also wouldn't `this->m_surface == m_surface`?

shell/egl_window.cc
```diff
@@ -336,6 +336,9 @@ void EglWindow::redraw(void* data,
   if (callback)
     wl_callback_destroy(callback);
 
+  window->m_callback = wl_surface_frame(window->m_surface);
+  wl_callback_add_listener(window->m_callback, &frame_listener, window);
```
mv0, line 373: I think this belongs in a different patch. Obviously not adding a new frame callback would not trigger it a new one. So this seems like a small bug missing adding it. Right?

shell/egl_window.cc
```diff
@@ -361,9 +364,9 @@ void EglWindow::toggle_fullscreen() {
     xdg_toplevel_unset_fullscreen(m_xdg_toplevel);
   }
 
-  //handle_xdg_surface_configure(this, m_xdg_surface, 0);
-//  wl_callback_add_listener(wl_display_sync(m_display->GetDisplay()),
-//                           &xdg_configure_callback_listener, this);
+  // handle_xdg_surface_configure(this, m_xdg_surface, 0);
+  //  wl_callback_add_listener(wl_display_sync(m_display->GetDisplay()),
+  //                           &xdg_configure_callback_listener, this);
```
mv0: Again, this probably is just indenting thing.

shell/egl_window.h
```diff
@@ -112,32 +113,10 @@ class EglWindow : public Egl {
...
-                                  int width,
-                                  int height,
-                                  uint32_t time);
-
-  static void paint_pixels(void* image,
-                           int padding,
-                           int width,
-                           int height,
-                           uint32_t time);
-
```
mv0, line 151: Clean-ups of these do belong in a different patch. These removal have nothing to do with patch. Or if so please why in the commit.

shell/egl_window.cc
```diff
@@ -49,8 +49,9 @@ EglWindow::EglWindow(size_t index,
   FML_DLOG(INFO) << "+ EglWindow()";
 
   m_surface = wl_compositor_create_surface(m_display->GetCompositor());
-  assert(m_surface);
-  FML_DLOG(INFO) << "EglWindow::m_surface = " << static_cast<void*>(m_surface);
+#if HIDPI_CLIENT
```
mv0: Why only for HDPI_CLIENT? 

What's the reasoning. There's no commit description explaining it.

shell/egl_window.cc
```diff
@@ -49,8 +49,9 @@ EglWindow::EglWindow(size_t index,
   FML_DLOG(INFO) << "+ EglWindow()";
 
   m_surface = wl_compositor_create_surface(m_display->GetCompositor());
-  assert(m_surface);
-  FML_DLOG(INFO) << "EglWindow::m_surface = " << static_cast<void*>(m_surface);
```
mv0, line 54: Removal of these changes require a different patch.

shell/egl_window.cc
```diff
@@ -59,15 +60,9 @@ EglWindow::EglWindow(size_t index,
 
   m_xdg_surface =
       xdg_wm_base_get_xdg_surface(m_display->GetXdgWmBase(), m_surface);
-  assert(m_xdg_surface);
-  FML_DLOG(INFO) << "EglWindow::m_xdg_surface = "
-                 << static_cast<void*>(m_xdg_surface);
```
mv0: Same. Removal of stuff require a different patch.

shell/egl_window.cc
```diff
@@ -59,15 +60,9 @@ EglWindow::EglWindow(size_t index,
...
-
   xdg_surface_add_listener(m_xdg_surface, &xdg_surface_listener, this);
-  //  xdg_surface_add_listener(m_xdg_surface, &xdg_surface_listener,this);
```
mv0: This belongs in a previous patch.

shell/egl_window.cc
```diff
@@ -59,15 +60,9 @@ EglWindow::EglWindow(size_t index,
...
   m_xdg_toplevel = xdg_surface_get_toplevel(m_xdg_surface);
   xdg_toplevel_set_title(m_xdg_toplevel, m_app_id.c_str());
-  //  wl_surface_commit(m_surface);
```
mv0: Belongs in a different/previous patch.

shell/egl_window.cc
```diff
@@ -84,18 +79,8 @@ EglWindow::EglWindow(size_t index,
 
   m_egl_surface[index] = create_egl_surface(this, m_egl_window[index], nullptr);
 
-  toggle_fullscreen();
```
mv0, line 85: Yeah, I've already made a comment about this, but reviewing happens patch by patch. I would introduce these changes in first patch probably.

shell/egl_window.cc
```diff
@@ -325,6 +310,30 @@ void EglWindow::handle_xdg_surface_configure(void* data,
   w->m_xdg_surface_is_configured = true;
 }
 
+#if HIDPI_CLIENT
```
mv0: Again, no explanation as to why only for HIDPI_CLIENT.

shell/display.cc
```diff
@@ -741,6 +715,53 @@ bool Display::ActivateSystemCursor([[maybe_unused]] int32_t device,
...
+    }
+  }
+}
+
+const struct wl_surface_listener Display::primary_surface_listener = {
+    handle_primary_surface_enter};
+
+void Display::SetSurface(wl_surface* surface) {
+  m_surface = surface;
+  // requires compositor v3+
```
mv0, line 764: What is compositor v3?

shell/display.cc
```diff
@@ -741,6 +715,53 @@ bool Display::ActivateSystemCursor([[maybe_unused]] int32_t device,
...
+  // wl_surface_set_buffer_scale(m_surface, m_buffer_scale);
```
mv0, line 765: Left over?

shell/display.h
```diff
@@ -177,33 +166,24 @@ class Display {
...
+    struct wl_output* output;
+    uint32_t global_id;
+    unsigned width;
+    unsigned height;
+    unsigned physical_width;
+    unsigned physical_height;
+    int refresh_rate;
+    int32_t scale;
+    bool done;
+  } output_info_t;
```
mv0, line 179: So... .wouldn't this better be a class? Instead of typedefing stuff. Its CPP after all.

shell/egl_window.cc
```diff
@@ -62,39 +56,29 @@ EglWindow::EglWindow(size_t index,
       xdg_wm_base_get_xdg_surface(m_display->GetXdgWmBase(), m_surface);
   xdg_surface_add_listener(m_xdg_surface, &xdg_surface_listener, this);
   m_xdg_toplevel = xdg_surface_get_toplevel(m_xdg_surface);
+  xdg_toplevel_add_listener(m_xdg_toplevel, &xdg_toplevel_listener, this);
```
mv0, line 59: This needs to be in the first patch.

shell/egl_window.cc
```diff
@@ -305,37 +289,77 @@ int EglWindow::create_shm_buffer(Display* display,
...
+  if (width > 0 && height > 0) {
+    if (!w->m_fullscreen && !w->m_maximized) {
+      w->m_window_size.width = width;
+      w->m_window_size.height = height;
+    }
+    w->m_geometry.width = width;
+    w->m_geometry.height = height;
+
+    /* Initialise damage to full surface, so the padding gets painted */
+    wl_surface_damage(w->m_surface, 0, 0, w->m_geometry.width,
```
mv0, line 337: No need to inflict damage for the surface.

shell/egl_window.cc
```diff
@@ -305,37 +289,77 @@ int EglWindow::create_shm_buffer(Display* display,
...
+                      w->m_geometry.height);
+
+    if (w->m_egl_window[w->m_index]) {
```
mv0, line 340: I would do the resize at the end.

shell/egl_window.cc
```diff
@@ -366,16 +390,8 @@ const struct wl_callback_listener EglWindow::frame_listener = {redraw};
...
-    xdg_toplevel_set_fullscreen(m_xdg_toplevel, m_display->GetOutput());
-  } else {
-    xdg_toplevel_unset_fullscreen(m_xdg_toplevel);
-  }
-
-  // handle_xdg_surface_configure(this, m_xdg_surface, 0);
-  //  wl_callback_add_listener(wl_display_sync(m_display->GetDisplay()),
-  //                           &xdg_configure_callback_listener, this);
+void EglWindow::set_fullscreen() {
+  xdg_toplevel_set_fullscreen(m_xdg_toplevel, nullptr);
```
mv0, line 394: You've changed this 3 times in this series. I would keep the bool as is. You might want to change from fullscreen to unfulscreen on different shells.

shell/egl_window.h
```diff
@@ -27,7 +27,13 @@
 #include "egl.h"
 #include "xdg-shell-client-protocol.h"
 
+#define WL_ARRAY_FOR_EACH(pos, array, type)                             \
```
mv0, line 30: #ifndef ....

Please also explain why we need this.... I'd like to see some cpp construct that fixes this... seems a bit weird that can't really cast generic pointers. In C we don't even need an explicit cast.

shell/egl_window.h
```diff
@@ -67,22 +72,30 @@ class EglWindow : public Egl {
...
   bool m_fullscreen;
-  bool m_sprawl;
+  bool m_maximized{};
+  bool m_resize{};
+  bool m_activated{};
+  bool m_running{};
+  struct {
+    int32_t width;
+    int32_t height;
+  } m_geometry;
```
mv0, line 88: `m_geometry, m_window_size{}`?

shell/egl_window.h
```diff
@@ -67,22 +72,30 @@ class EglWindow : public Egl {
...
+  struct {
+    int32_t width;
+    int32_t height;
+  } m_window_size{};
+
```
mv0, line 93: No need do declare once more, there's one above.

shell/egl_window.h
```diff
@@ -94,7 +107,7 @@ class EglWindow : public Egl {
 
   int m_frame_sync;
 
-  void set_fullscreen(bool val);
+  void set_fullscreen();
```
mv0, line 110: Wasn't okay to able to go back from fullscreen to unfullscreen?

shell/egl_window.h
```diff
@@ -107,25 +120,22 @@ class EglWindow : public Egl {
...
+                                        struct xdg_toplevel* toplevel,
+                                        int32_t width,
+                                        int32_t height,
+                                        struct wl_array* states);
 
-  static void handle_primary_surface_leave(void* data,
-                                           struct wl_surface* wl_surface,
-                                           struct wl_output* output);
-#endif
+  static void handle_toplevel_close(void* data,
```
mv0, line 137: All of the toplevel bits need to go in the first patch, added here need to go in the first patch.

shell/engine.h
```diff
@@ -122,9 +122,9 @@ class Engine {
   std::string GetAssetDirectory() { return m_assets_path; }
 
 #if ENABLE_PLUGIN_TEXT_INPUT
-  TextInput* m_text_input;
+  TextInput* m_text_input{};
   [[maybe_unused]] void SetTextInput(TextInput *text_input);
-  [[maybe_unused]] TextInput *GetTextInput();
+  [[maybe_unused]] TextInput *GetTextInput() const;
```
mv0, line 127: These belong in a different patch.

shell/egl_window.cc
```diff
@@ -62,39 +56,29 @@ EglWindow::EglWindow(size_t index,
...
-  FML_DLOG(INFO) << "create egl_window: " << m_width << "x" << m_height;
-
-  m_egl_surface[index] = create_egl_surface(this, m_egl_window[index], nullptr);
-
-  set_fullscreen(m_fullscreen);
-
-  /* Initialise damage to full surface, so the padding gets painted */
-  wl_surface_damage(m_surface, 0, 0, m_width, m_height);
+  m_egl_window[m_index] =
+      wl_egl_window_create(m_surface, m_geometry.width, m_geometry.height);
```
mv0, line 66: The m_geometry.width and m_geometry.height are passed from App() right? 

In practice, are these the updated values from the resize  in top level configure or the initial ones?

Have you tried these in system with software rendering?

mv0:
My initial comments for these are:

- please keep the cleanups in a different patch/patches
- keep together the wl_shell to xdg-shell conversion - do not split toplevel configure events from xdg_surface ones. Makes reviewing much easier
- all patches are missing a proper commit description

shell/egl_window.cc
jwinarske: The initial values passed from App() get updated via handle_toplevel_configure().

I have not yet tried this PR with a software render (Qemu).  I need to validate it on my CI platforms (meta-flutter).

I am planning at some point to add optional support for a decoration library for a more friendly desktop support.  So resizing etc will be added later.  The primary use case is a minimal target app, that happens to work on desktop.

shell/display.h
```diff
@@ -237,7 +238,7 @@ class Display {
                                    int scale);
   static void display_handle_done(void* data, struct wl_output* wl_output);
 
-  bool m_is_configured;
+  bool m_is_configured{};
```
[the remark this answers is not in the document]
jwinarske: Yes.

shell/egl_window.cc
```diff
@@ -80,21 +84,17 @@ EglWindow::EglWindow(size_t index,
 
   m_egl_surface[index] = create_egl_surface(this, m_egl_window[index], nullptr);
 
-  wl_shell_surface_set_title(m_shell_surface, m_app_id.c_str());
-
   toggle_fullscreen();
```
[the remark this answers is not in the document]
jwinarske: Moved this to !m_fullscreen case.

shell/egl_window.cc
```diff
@@ -317,122 +317,16 @@ int EglWindow::create_shm_buffer(Display* display,
...
-
-void EglWindow::paint_pixels_bottom(void* image,
-                                    [[maybe_unused]] int padding,
-                                    int width,
-                                    int height,
-                                    [[maybe_unused]] uint32_t time) {
-  memset(image, 0xa0,
-         static_cast<size_t>(width) * static_cast<size_t>(height) * 4);
+  xdg_surface_ack_configure(xdg_surface, serial);
+  w->m_xdg_surface_is_configured = true;
```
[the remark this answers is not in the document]
jwinarske: The wl_callback_add_listener is used for painting the FPS window.

shell/egl_window.h
```diff
@@ -72,7 +69,7 @@ class EglWindow : public Egl {
   std::shared_ptr<Display> m_display;
 
   struct wl_surface* m_surface;
-  struct wl_shell_surface* m_shell_surface;
+  struct wl_shell_surface* m_shell_surface{};
```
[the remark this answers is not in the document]
jwinarske: yep

shell/egl_window.cc
jwinarske: yes, left over from prior patch.  remove

shell/egl_window.h
jwinarske: The base Wayland macro does not compile as-is in C++.  This is pretty well documented on the internet.  I added a comment in updated patches.

shell/egl_window.h
jwinarske: Optional window decorations are next hence window vs geometry.

shell/egl_window.cc
jwinarske: removed

shell/egl_window.cc
jwinarske: this is for rendering the FPS overlay window

shell/display.cc
jwinarske: noted and addressed

shell/display.h
jwinarske: If you search the source tree you'll find many references to this pattern.  The goal is to stay as close to Google/Flutter standards.  This makes upstreaming much easier.  So this is fine.

shell/egl_window.cc
jwinarske: moved.

shell/egl_window.cc
```diff
@@ -52,18 +51,23 @@ EglWindow::EglWindow(size_t index,
...
+  m_xdg_surface =
+      xdg_wm_base_get_xdg_surface(m_display->GetXdgWmBase(), m_surface);
+  assert(m_xdg_surface);
+  FML_DLOG(INFO) << "EglWindow::m_xdg_surface = "
+                 << static_cast<void*>(m_xdg_surface);
+
+  xdg_surface_add_listener(m_xdg_surface, &xdg_surface_listener, this);
+  //  xdg_surface_add_listener(m_xdg_surface, &xdg_surface_listener,this);
+  m_xdg_toplevel = xdg_surface_get_toplevel(m_xdg_surface);
+  xdg_toplevel_set_title(m_xdg_toplevel, m_app_id.c_str());
```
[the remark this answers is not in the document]
jwinarske: done.

jwinarske:
Yeah I agree.  This wasn't ready for review when I pointed it out.  I refactored the patches for logical grouping.

shell/egl_window.h
jwinarske: https://gitlab.freedesktop.org/wayland/wayland/-/issues/34

cmc5788 (dismissed):

shell/egl_window.h
mv0: Yeah, I was just hinting that we can use the same struct  here instead of declaring a new anonymous one.

mv0:
Had a few minor nitpicks, these now look fine to me now. I (still) need to do a follow-up with the resize, as that will currently will be an issue when starting up in a different state (max, rather than full). Feel free to merge once resolved (the comments).

README.md
```diff
@@ -108,6 +108,21 @@ Copy libflutter_engine.so to `/usr/local/lib` or use LD_LIBRARY_PATH to point do
     export LD_LIBRARY_PATH=`pwd`:$LD_LIBRARY_PATH
     homescreen
 
+## FPS
+Three environmental variables impact FPS reporting.
```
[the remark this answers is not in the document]
jwinarske: no I haven't.  I'll check that out.

shell/display.cc
```diff
@@ -194,69 +198,55 @@ void Display::display_handle_geometry(void* data,
...
-    FML_DLOG(INFO) << "subpixel: " << d->m_info.geometry.subpixel;
-    FML_DLOG(INFO) << "make: " << d->m_info.geometry.make;
-    FML_DLOG(INFO) << "model: " << d->m_info.geometry.model;
-    FML_DLOG(INFO) << "transform: " << d->m_info.geometry.transform;
-  }
+  auto* oi = static_cast<output_info_t*>(data);
+  oi->physical_width = physical_width;
+  oi->physical_height = physical_height;
+
+  FML_DLOG(INFO) << "Physical width: " << physical_width << " mm x "
```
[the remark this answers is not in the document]
jwinarske: Correct.  Pass `-DCMAKE_BUILD_TYPE=Debug` to CMake.

shell/app.cc
```diff
@@ -107,8 +107,9 @@ App::App(const std::string& app_id,
   m_fps_counter = 0;
 
   const char* envstr_console;
+  char* pEnd;
   if ((envstr_console = getenv("FPS_OUTPUT_CONSOLE")) != nullptr) {
-    int val = atoi(envstr_console);
+    long val = strtol(envstr_console, &pEnd, 10);
```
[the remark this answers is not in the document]
jwinarske: good point.  removed pEnd for nullptr.

shell/display.cc
```diff
@@ -32,15 +33,16 @@ Display::Display([[maybe_unused]] App* app,
...
       m_subcompositor(nullptr),
       m_cursor_surface(nullptr),
       m_keyboard(nullptr),
       m_xkb_context(xkb_context_new(XKB_CONTEXT_NO_FLAGS)),
       m_keymap(nullptr),
       m_xkb_state(nullptr),
-      m_agl_shell(nullptr),
+      m_buffer_scale(1),
+      m_last_buffer_scale(1),
+	  m_agl_shell(nullptr),
```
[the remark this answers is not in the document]
jwinarske: I've refactored this.  All nullptr initialization is happening at point of definition.

shell/app.cc
```diff
@@ -57,7 +57,7 @@ App::App(const std::string& app_id,
 
   FML_DLOG(INFO) << "+App::App";
 
-  m_display->AglShellDoBackground(m_egl_window[0]->GetNativeSurface());
+  m_display->AglShellDoBackground(m_egl_window[0]->GetNativeSurface(), 0);
```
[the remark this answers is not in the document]
jwinarske: Yeah I agree.  I'll add `constexpr int` values to constants.h once the layout is logically tied to a use case.  For now it's more of a placeholder than anything.

shell/egl_window.cc
```diff
@@ -53,51 +52,35 @@ EglWindow::EglWindow(size_t index,
...
-  this->m_callback = wl_surface_frame(this->m_surface);
-  wl_callback_add_listener(this->m_callback, &frame_listener, this);
-  wl_surface_commit(this->m_surface);
+  m_callback = wl_surface_frame(m_surface);
+  wl_callback_add_listener(m_callback, &frame_listener, this);
+
+  if (m_fullscreen)
+    xdg_toplevel_set_fullscreen(m_xdg_toplevel, nullptr);
+  else
+    xdg_toplevel_set_title(m_xdg_toplevel, m_app_id.c_str());
```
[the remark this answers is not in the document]
jwinarske: I moved it back.

shell/display.cc
```diff
@@ -194,69 +198,55 @@ void Display::display_handle_geometry(void* data,
...
-    FML_DLOG(INFO) << "scale: " << d->m_info.scale.scale;
-  }
+  FML_DLOG(INFO) << "Display Scale Factor: " << factor;
 }
 
 void Display::display_handle_done(void* data, struct wl_output* wl_output) {
-  (void)data;
   (void)wl_output;
+  auto* oi = static_cast<output_info_t*>(data);
+  oi->done = true;
```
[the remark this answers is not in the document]
jwinarske: Hmm.  On multiple platforms I'm seeing done always show up last.  I'm not currently being used for anything.

shell/display.h
```diff
@@ -166,30 +168,24 @@ class Display {
...
+  typedef struct output_info {
+    struct wl_output* output;
+    uint32_t global_id;
+    unsigned width;
+    unsigned height;
+    unsigned physical_width;
+    unsigned physical_height;
+    int refresh_rate;
+    int32_t scale;
+    bool done;
```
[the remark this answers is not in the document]
jwinarske: I need to read the spec on this.  I see done in the log as the last wl_output event.  Again I'm not taking action on the value.

shell/display.h
```diff
@@ -166,30 +168,24 @@ class Display {
...
-    } scale{};
-
-  } m_info;
+  typedef struct output_info {
+    struct wl_output* output;
+    uint32_t global_id;
+    unsigned width;
+    unsigned height;
+    unsigned physical_width;
+    unsigned physical_height;
```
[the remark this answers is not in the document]
jwinarske: Yeah I'm not using the EDID description at this time.  I may add this back to pass to upper level app at some point.  There are other things that need to be reported as well.  Not a firm use case yet.

shell/display.cc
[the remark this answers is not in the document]
jwinarske: not sure what you're referring to here.

jwinarske:
@mv0 I have updates that handle buffering scaling, and setup interface versioning using std::min logic.  I've moved forward on wl_seat, etc.  I can now drag the window between screens and the buffer scaling changes.  It's not to say the Mutter implementation tracks 100% correctly, but the nuance can be addressed at a later date if it shows up else where.  There was an issue when display scaling was > 100% it would segfault; without these changes.

Issue #56, opened 2022-06-22
Mouse pointer motion events sometimes violate debug assertion check for state.is_down


doesnotexist:
On my ubuntu 20.04, wayland+weston+ivi-homescreen running the default flutter create "increment" button app with debug engine v2.10.3 arm64,  I'm finding that the app often crashes when using mouse input because this assertion check fails.

```
[FATAL:flutter/lib/ui/window/pointer_data_packet_converter.cc(189)] Check failed: !state.is_down. 
```

https://github.com/flutter/engine/blob/main/lib/ui/window/pointer_data_packet_converter.cc#L190

I believe this is a manifestation of a bug in translating wayland input events to flutter, specifically within Display::pointer_handle_motion()
https://github.com/toyota-connected/ivi-homescreen/blob/main/shell/display.cc#L399

Which calls getPointerPhase(), and in some instances the state can be such that p->buttons is not null and state == WL_POINTER_BUTTON_STATE_RELEASED which ends up translating the event to FlutterPointerPhase::kUp.
https://github.com/toyota-connected/ivi-homescreen/blob/main/shell/display.cc#L337
However, if no kDown event is sent between this kUp event and some previous kUp event (say from releasing the mouse button) then the debug assert fails and the app crashes with the FATAL warning above.

Based on what I've gathered from reading the comments in embedder.h
https://github.com/flutter/engine/blob/main/shell/platform/embedder/embedder.h#L727-L743

I propose that the wayland input event listener Display::pointer_handle_motion()  should only translate to  FlutterPointerPhase::kMove or  FlutterPointerPhase::kHover  depending on whether the left mouse button is currently pressed down or not during the motion event.   PR that implements this change coming shortly.

doesnotexist:
This bug doesn't happen in touch handlers because a boolean based on counter down_count prevents multiple kUp events from being ever being sent to flutter. 
https://github.com/toyota-connected/ivi-homescreen/blob/main/shell/display.cc#L636

doesnotexist:
The linux desktop shell implements what I proposed above: https://github.com/flutter/engine/blob/main/shell/platform/linux/fl_view.cc#L383

jwinarske:
PR merged

PR #57, opened 2022-06-22
Simplify wayland mouse input translation

doesnotexist:
Addressing issue described in #56  and simplifying the previous wayland mouse input events to flutter pointer events translation in ivi-homescreen:

Sanity checked implementation here against the official flutter linux shell's mouse input handling for motion, scrolling and button events. The relevant analogous code sections are linked below:

Motion:
https://github.com/flutter/engine/blob/main/shell/platform/linux/fl_view.cc#L371-L372
motion events are either kMove or KHover, matching linux desktop:
https://github.com/flutter/engine/blob/main/shell/platform/linux/fl_view.cc#L383

Scrolling (Axis in wayland):
https://github.com/flutter/engine/blob/main/shell/platform/linux/fl_scrolling_manager.cc#L67
Notice that for scrolling events the pointer phase is disregarded and kMove is chosen arbitrarily based on this comment
https://github.com/flutter/engine/blob/main/shell/platform/linux/fl_scrolling_manager.cc#L122-L123

Button:
https://github.com/flutter/engine/blob/main/shell/platform/linux/fl_view.cc#L114-L115
Currently this PR does not drop redundant events but perhaps it should try to mimic such behavior.  Another difference is that this PR does not keep track of previous button state and so only sends kUp or kDown events for button events, never kMove or KHover, it's unclear to me if this should behavior should be copied or not.
https://github.com/flutter/engine/blob/main/shell/platform/linux/fl_view.cc#L130-L148

Tested this PR manually on the default flutter increment button app which no longer crashes with debug assertions described in Issue #56.

### DIFF
--- a/shell/display.cc
+++ b/shell/display.cc
@@ -318,32 +318,8 @@ const struct wl_seat_listener Display::seat_listener = {
     .name = seat_handle_name,
 };
 
-FlutterPointerPhase Display::getPointerPhase(struct pointer* p) {
-  [[maybe_unused]] FlutterPointerPhase res = p->phase;
-
-  if (p->buttons) {
-    switch (p->event.state) {
-      case WL_POINTER_BUTTON_STATE_PRESSED:
-        // FML_DLOG(INFO) << "WL_POINTER_BUTTON_STATE_PRESSED";
-        if ((p->phase == FlutterPointerPhase::kHover) ||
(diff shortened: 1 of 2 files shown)


jwinarske:
I just confirmed it on Gallery.  Fix looks good.  Thanks for the contribution!

PR #59, opened 2022-06-23
Enable pixel_ratio arg from CLI

cmc5788:
Allows making `pixel_ratio` something other than `1.0` using optional command line arg.

Examples --

```shell
./homescreen --w=640 --h=480 --r=2.0
```

Screen Shot 2022-06-22 at 8 44 03 PM

```shell
./homescreen --w=640 --h=480 --r=0.5
```

Screen Shot 2022-06-22 at 8 44 21 PM

### DIFF
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -33,6 +33,7 @@ App::App(const std::string& app_id,
          bool debug_egl,
          uint32_t width,
          uint32_t height,
+         double pixel_ratio,
          const std::string& cursor_theme_name)
     : m_gl_resolver(std::make_shared<GlResolver>()),
       m_display(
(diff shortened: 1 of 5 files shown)


jwinarske:
@cmc5788 
1. I have had PixelRatio auto calculation change pending.  I had it as part of xdg_shell rework, and was holding till that stabilized.  It's now ready to go in.
2. After this, there could be a Pixel Ratio scale factor parameter, which was multiplied with the auto calculated pixel ratio.  I would only enable this in debug builds.

cmc5788:
@jwinarske -- okay, let's see what happens after that change, but we'll probably still need the ability to set the scale factor in production builds regardless of what the display calculates. We need to be able to control the number of `logical pixels` reported by Flutter regardless of what display reports. Not only for debug.

PR #71, opened 2022-07-22
Allow the embedder to run as a regular/normal application

mv0:
These changes allows the embedder, on the AGL platforms/systems, to run 
as a regular/normal application. 

Specifically, the changes here do the following:

- avoid a bind to agl-shell if the compositor advertises it. Without a protocol change this won't be easily fixed, but the
client can infer easily from the configuration/cmd line if it needs to actually bind to the interface. Plan on addressing that in the compositor but that would take some time. When I have done that we can revisit this change. 
- do not issue always the ready request if we didn't found a bg type of window. While technically there's nothing that stop
on doing it, it won't really work. Not setting a background surface (even with setting a panel) won't really work as we can not determine clients geometry nor can we deduce the panel dimensions (depending on the orientation). Further more, default
activation does require having at least background surface. 

I kinda in-lined these changes so probably they need a bit more massaging.

### DIFF
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -25,7 +25,8 @@
 
 App::App(const std::vector<Configuration::Config>& configs)
     : m_wayland_display(std::make_shared<Display>(!configs[0].disable_cursor,
-                                                  configs[0].cursor_theme)) {
+                                                  configs[0].cursor_theme,
+                                                  configs)) {
   FML_DLOG(INFO) << "+App::App";
   bool found_view_with_bg = false;
 
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -27,6 +27,7 @@ App::App(const std::vector<Configuration::Config>& configs)
     : m_wayland_display(std::make_shared<Display>(!configs[0].disable_cursor,
                                                   configs[0].cursor_theme)) {
   FML_DLOG(INFO) << "+App::App";
+  bool found_view_with_bg = false;
 
   size_t index = 0;
   m_views.reserve(configs.size());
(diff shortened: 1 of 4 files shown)


jwinarske (approved):

shell/app.cc
```diff
@@ -35,9 +37,16 @@ App::App(const std::vector<Configuration::Config>& configs)
     view->Initialize();
     m_views.emplace_back(std::move(view));
     index++;
+
+    if (WaylandWindow::get_window_type(cfg.view.window_type) ==
```
jwinarske, line 41: I think this is in conflict with my recent change.  I make BG the default window type.  The reasoning is so that one can run cli options not including json config, as json config was only way to set window type.

mv0: Yeah, pretty much. 

> The reasoning is so that one can run cli options not including json config

Not really sure how to reconcile that. To me this sounds like a reason to keep it as normal, as the default, while for background/panel being enabled if either a) explicitly specified over config file, or b) passed over as cmd line. 

This way:

- for non-AGL platforms, we resort to using normal role (and the agl calls are stubs as it won't find the protocol interface being advertised)
- on AGL platforms where the flutter platform is not the default shell client, it can run flutter as regular/normal applications
- on AGL platforms where the flutter platforms is the shell client, you explicitly state it (and avoid a potential conflict with other running flutter apps). 

At some point in time, I'll update the protocol interface such that the client can still bind to the protocol interface and determine if there's another shell client running. With that in place, the client can infer what do do next (it lost the race, there's another shell client running, or not). Until that happens, adding an explicit command line argument or using the config file seems to be the easiest to go forward.

jwinarske: I think the way to reconcile is pass widow type as cli option.  Then revert default back to `normal`.  I'll create a PR for this.  I just wanted to unblock the need to pass --j=

mv0: Well then I guess with this change we break again functionality of ivi-homescreen as our applications will need to pass that argument.

PR #73, opened 2022-07-25
Minor parsing fixes for JSON configuration file

mv0:
This patch series fixes some minor parsing isssues I've found while using the JSON config file.

Specifically, app_id specified over the configuration is being ignored and found similar issue with debug_backend and disable_cursor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,7 +175,7 @@ Loads Single View
 
 {
    "disable_cursor":true,
-   "backend_debug":true,
+   "debug_backend":true,
    "accessibility_features":31,
    "view":{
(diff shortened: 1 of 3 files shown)


jwinarske:
updated README.md in `main` with backend_debug fix.

mv0:
Any reason for closing this PR?

jwinarske:
@mv0 
1. Application id is global, and not unique per view.  Only index 0 should be used, as it will go away with multi-display feature.
2. `if (cli.debug_backend != instance.debug_backend)` is correct.  If the value is different than stored, then assign it.  The struct has a default value of false.  So only if debug_backend is true will it be stored as such.
3. I updated README.md in `main` with the typo.

PR #74, opened 2022-07-28
Change Default Window Type

jwinarske:
-enables default window on AGL as Background
 prevents needing json config
-harmless in desktop case, as it checks if agl_shell is not nullptr


mv0:
I think is is a mistake, defaulting to background surface role would make running ivi-homescreen impossible if running as as simple application. 

With this change we basically break functionality. 

This pretty backwards to provide an argument to use default values.

jwinarske:
@mv0 This is required to enable using CLI to run application on AGL without requiring `--j=` option.  The fix will be to add CLI option to select window type.

PR #88, opened 2023-01-04
Handle selectMultiEntryHistory message in nativation plugin

timbz:
This is needed for [go_router](https://pub.dev/packages/go_router) to function


jwinarske:
@timbz Thanks Timon!

PR #109, opened 2023-06-01
Couple of fixes related to transform orientation and width/height 

mv0:
These fix some trivial issues with wl_output events. The orientation one is of particular useful to have. For the width and height without having output_info_t as a list we will be overriding the width and height for each display_handle_mode event we get. 

Both these seem unused at this point but we'll make use of them in AGL in order to have an activation area being set-up irrespective of the output dimensions or its orientation.


jwinarske:
Please rebase to the AGL branch

jwinarske:
Closing since `main` is for publishing from internal Toyota repo, and same change was just merged into `agl` branch.

PR #112, opened 2023-06-19
shell/wayland/display: Handle flags as bitfield entries

mv0:
We can have preferred and current mode or'ed in the same time, so we can't really check just for current mode. Make sure we do get a valid width/height for the outputs to avoid going further.

Bug-AGL: SPEC-4841

### DIFF
--- a/shell/wayland/display.cc
+++ b/shell/wayland/display.cc
@@ -258,11 +258,13 @@ void Display::display_handle_mode(void* data,
   (void)flags;
   auto* oi = static_cast<output_info_t*>(data);
 
-  if (flags == WL_OUTPUT_MODE_CURRENT) {
+  if ((flags & WL_OUTPUT_MODE_CURRENT) == WL_OUTPUT_MODE_CURRENT) {
     oi->height = static_cast<unsigned int>(height);
     oi->width = static_cast<unsigned int>(width);
     oi->refresh_rate = refresh;
   }
+
+  assert(oi->height && oi->width);
 
   FML_DLOG(INFO) << "Video mode: " << width << " x " << height << " @ "
                  << (refresh > 1000 ? refresh / 1000.0 : (double)refresh)


jwinarske:
I'll take as look at this tomorrow. 

Also I need to confirm if we have a regression on multi output.  I use the output enumeration to stash the output pointer, which is used to assign Fullscreen output using index from configuration.

mv0:
> Also I need to confirm if we have a regression on multi output. I use the output enumeration to stash the output pointer, which is used to assign Fullscreen output using index from configuration.

Thanks a lot! I used to think this might be issue, picking the wrong output. Turns out I just ignored some really common use case. I mean now that you mention it, I would spend/dedicate some time coming up with a better system to manage outputs, rather than yeah, indexing things into an array. 

wl_output primitive is used extensively as a way to move underlying compositor data structures between different outputs. 

Probably some kind of List abstraction on top of the array, or well baking something else entirely might be more suitable. Grabbing the first/last/getters by name.description, would fare better than just integer numbers. 

Also small fyi, with further updates to the agl-shell protocol, it gives the shell client a bit more leeway to choose which output to move/migrate/start application on different outputs. So this abstraction itself would come in hand when well doing another update for the agl-shell protocol.

mv0:
The client side in weston is handled by clients/window.*, under a the name library called toytoolkit. Most of the simple-* clients are using that library. We usually tend to steer away from that, but that's where we keep the list of outputs. Recently wl_output interface in wayland got some updates which made it through some of the changes that were added prior in a protocol extension named xdg_output_manager. Weston itself as a compositor is capable of managing all heads and outputs, there's nothing limiting it to a single output. Depending on the shell, there's a different policy: randomly place windows and pick based on the pointer is, like desktop-shell, or like kiosk-shell where we can specify which outputs to place the application based on the app_id. ivi-shell on its own is a different thing, makes use of the controller to do that. 

I've mostly working with Qt, so they have something call a QScreen which abstracts the entire output under that, irrespective of the windowing system. There's a bridge between that and the actually system, like for instance qtwayland which uses something called a  QWaylandScreen. I don't think you need this second level of abstraction given that you explicitly target Wayland. I would model that after the protocol interface (https://wayland.app/protocols/wayland#wl_output) and keep a list of each object part of Display. 

Mutter, KWIN mostly likely have something similar.

jwinarske:
I see.  That was helpful thanks.  I think the challenge remains on how to associate a unique key to a given output.  Say you start with a known list pattern, if you unplug, or hotplug another things become out of order.  Then next restart with all plugged in, the order is different.  My initial implementation takes the order of occurrence at start as that's when configuration is applied.  How can I determine the place order of a hot plug monitor on a subsequent Weston restart?  Say a user powers off the HUD, then later powers it back on during runtime.

mv0:
> I think the challenge remains on how to associate a unique key to a given output. Say you start with a known list pattern, if you unplug, or hotplug another things become out of order

Hmm, why would they become out of order? What exactly becomes out of order? What are the assumptions here? Do you have multiple clients each covering each output? Do you have a single client covering both outputs? 

We get events when on wl_output comes in and out of existence. This is what we as wayland clients get to work with. https://wayland.app/protocols/xdg-output-unstable-v1#zxdg_output_v1:event:logical_position might be useful to correctly identify the order in the compositor space. Haven't used it, so I can't tell much about it but it might seem useful.  

> Then next restart with all plugged in, the order is different.

Think this is something outside of the compositor control, being an issue that can be solved by using udev rules, rather than trying to overcome it with code. 

For instance, having 4 outputs each with an panel/touch input device the following would always map input and outputs in the same way:

```
SUBSYSTEM=="input", ATTRS{idVendor}=="222a", ATTRS{idProduct}=="004a", OWNER="display", ENV{ID_SEAT}="seat0", ENV{WL_OUTPUT}="HDMI-A-1"
SUBSYSTEM=="input", ATTRS{idVendor}=="222a", ATTRS{idProduct}=="004b", OWNER="display", ENV{ID_SEAT}="seat0", ENV{WL_OUTPUT}="HDMI-A-2"
SUBSYSTEM=="input", ATTRS{idVendor}=="222a", ATTRS{idProduct}=="004c", OWNER="display", ENV{ID_SEAT}="seat0", ENV{WL_OUTPUT}="HDMI-A-3"
SUBSYSTEM=="input", ATTRS{idVendor}=="222a", ATTRS{idProduct}=="004d", OWNER="display", ENV{ID_SEAT}="seat0", ENV{WL_OUTPUT}="HDMI-A-4"
```

Fyi, in practice, hot-plugging a connector shows as being the same output, but I guess we'd like to be sure.

jwinarske:
Ah!  That makes a whole lot more sense.  That's what I was missing.  Cool thanks!

PR #127, opened 2023-11-02
backend/wayland_egl: Update damage area

mv0:
Rather than re-using the initial values from the config, update the values we are getting from the compositor to match the windows' dimensions.

This fixes some artifacts we're seeing when re-drawing with swap buffer with damage and most likely would have caused issues on other compositors when resizing the window to a higher value that the one provided in the configuration file.

Bug-AGL: SPEC-4938


jwinarske (approved):

jwinarske:
LGTM.  I merged to our internal repo, and cherry picked to main.

PR #128, opened 2023-11-19
use BUILD_DOC option

aduskett:
Do not build documentation if the BUILD_DOC option is set to OFF.


jwinarske:
I'm fine with this.  Main is downstream, so I will need to patch internal then push out. I'll pick this up in my next batch of changes.

jwinarske:
[2996a3164a5856cba882a8e8f495a74801085e59](https://github.com/toyota-connected/ivi-homescreen/commit/2996a3164a5856cba882a8e8f495a74801085e59)

PR #129, opened 2023-12-01
fix building with crash handling enabled

aduskett:
Two fixes needed:
  - The space after "CRASH_HANDLER_RELEASE=" results in the following error:
     """ ./constants.h:86:69: error: expected primary-expression before ‘;’ token 86 | static constexpr char 
           kCrashHandlerRelease[] = CRASH_HANDLER_RELEASE; 
     """
Remove the space

  - homescreen links against libunwind when BUILD_CRASH_HANDLER is set, however libunwind is not searched for when configuring. Use pkg_check_modules as libunwind does not install a .cmake file.

### DIFF
--- a/cmake/options.cmake
+++ b/cmake/options.cmake
@@ -113,10 +113,12 @@ endif ()
 #
 option(BUILD_CRASH_HANDLER "Build Crash Handler" OFF)
 if (BUILD_CRASH_HANDLER)
+    find_package(PkgConfig)
+    pkg_check_modules(LIBUNWIND REQUIRED libunwind)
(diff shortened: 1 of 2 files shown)


jwinarske:
I'm traveling this week.  I can get to this early next week

jwinarske:
Also heads up I will be building this out to be multi-platform in the next month:
https://github.com/toyota-connected/flutter-desktop
It's for dev work flow use cases (Mac etc)

jwinarske:
fixed in 5ab78a19e95c88cc5d6b173ab1260a211e78cf0a

PR #142, opened 2024-10-13
Update README.md to include recurse-submodules

Modd3d:
added --recurse-submodules -j8 to fix build cmake error when looking for the lists in the 3rd party repos and speed up cloning.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,14 +391,14 @@ Defaults to Wayland, no need to do anything special
 
 ### GCC/libstdc++ Build
 
-    git clone https://github.com/toyota-connected-na/ivi-homescreen.git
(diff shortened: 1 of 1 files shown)


jwinarske:
Also add section the show use of `git submodule update --init --recursive` if repo is already fetched.

Also I intentionally left out building with ivi-homescreen-plugins.  This might be a useful addition ;)

jwinarske:
also commits need to be squashed and signed

jwinarske:
resolved with https://github.com/toyota-connected/ivi-homescreen/commit/cb60bc8f27f461909cf3f9b3c15f54283d2296e1

PR #145, opened 2024-11-09
Refactor plugin initialization

jwinarske:
-moves plugin specific code to ivi-homescreen-plugins


jwinarske:
used for testing.  PR merged upstream

PR #148, opened 2025-05-27
Use pkgconfig for EGL and GLES libraries

g-scott-murray:
Mali GPU platforms that do not use Mesa have EGL and GLES implemented via libmali instead of the typical libEGL, etc.  Using pkgconfig for the link library options for those libraries allows that to work transparently when building the wayland backend.

### DIFF
--- a/shell/CMakeLists.txt
+++ b/shell/CMakeLists.txt
@@ -88,7 +88,9 @@ target_include_directories(${PROJECT_NAME}
 )
 
 if (BUILD_BACKEND_WAYLAND_EGL)
-    target_link_libraries(${PROJECT_NAME} PRIVATE EGL GLESv2)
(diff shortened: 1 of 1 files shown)


g-scott-murray:
The main driver for this for me is supporting Rockhip boards with the latest meta-rockchip (Rockchip's version), but pkgconfig for this libraries does seem like a reasonable expectation in general.  So far I've not seen any issues after testing this in the AGL build with Mesa using platforms like RPi, and non-Mesa platforms like Renesas and TI (both PowerVR).

jwinarske:
Makes sense.  Patch merged upstream
https://github.com/toyota-connected/ivi-homescreen/commit/2ca279f3bb6364da5c401450dc229a6584ff8514

PR #149, opened 2025-06-16
Fix AGL shell activation area behavior

g-scott-murray:
Commit cfc5de68 accidentally broke using an activation area configuration with either x or y set to 0.  Rework the logic to use the specified configuration area if any of the position or size values are non-zero.  This restores expected behavior in the downstream AGL demos.  As well, drop the fallback default activation area configuration, as it seems better to not configure it at all if it has not been explicitly configured.

### DIFF
--- a/shell/wayland/window.cc
+++ b/shell/wayland/window.cc
@@ -99,13 +99,11 @@ WaylandWindow::WaylandWindow(const size_t index,
       break;
     case WINDOW_BG:
       m_display->AglShellDoBackground(m_base_surface, 0);
-      if (m_activation_area.x && m_activation_area.y)
+      if (m_activation_area.x || m_activation_area.y ||
+	  m_activation_area.width || m_activation_area.height)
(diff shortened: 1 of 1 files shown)


jwinarske:
resolved with https://github.com/toyota-connected/ivi-homescreen/commit/7729fb5ca0d9021255321ff5c0904be5423893ba

g-scott-murray:
This change has nothing to do with the ENABLE_AGL_SHELL_CLIENT stuff we discussed, it's an actual logic fix.  We will still have to carry a patch for this in AGL to avoid a regression in application launching if this or an equivalent is not merged.

jwinarske:
ah.  sorry didn't read it!  Patching up stream

jwinarske:
fixed in https://github.com/toyota-connected/ivi-homescreen/commit/bc85acbad58b61da5fbfa97926c267ddb8e07abc

PR #155, opened 2025-10-01
Disabled PIE when linking

kerberjg:
Disables link-time `-pie` and mitigates `libzstd.a` not linking in `Release` mode


jwinarske:
We can’t disable this.  It’s mandated by the security team.  Seems we should set it when building filament-debug.

kerberjg:
Got it, I'll try to enable PIC compilation on libzstd somehow

PR #165, opened 2026-01-23
Fix Cmake with LLVM_ENABLE_RTTI_RAW disabled

kerberjg:
Syntax error fix: variable LLVM_ENABLE_RTTI_RAW is not being properly aliased in Cmake configuration, causing it to fail when empty.


kerberjg:
@jwinarske Ready for review ✨

jwinarske:
cc @kerberjg

PR #169, opened 2026-02-24
fix: remove non-async-signal-safe exit(0) from SIGINT handler

aki1770-del:
## Summary

Remove the `exit(0)` call from `SignalHandler()` in `shell/main.cc`.

## Problem

The SIGINT signal handler in `shell/main.cc` calls `exit(0)` immediately
after setting `running = false`. This has two issues:

1. **`exit()` is not async-signal-safe** (IEEE Std 1003.1-2017 §2.4.3).
   Calling it from a signal handler can cause undefined behavior if the
   signal interrupts a library function holding a lock.

2. **The `exit(0)` call makes `running = false` dead code** — the process
   terminates before the main loop can check the flag and perform graceful
   shutdown.

## Fix

Removing the `exit(0)` call allows the main loop to detect `running == false`
on its next iteration and exit naturally, ensuring:

- `App` destructor runs (`StopEvents()` on Wayland display)
- Logger cleanup completes (`spdlog::shutdown()`)
- Stack objects are properly destroyed

The signal handler becomes fully POSIX-compliant — it only calls
`SPDLOG_INFO` (which uses only async-signal-safe operations in this
context) and sets a `volatile sig_atomic_t`-compatible flag.

## Verification

- Production build: ✅ 2/2 targets (recompiled `main.cc.o` + re-linked `homescreen`)
- Environment-independent tests: ✅ 3/3 PASS (`utils`, `timer`, `gl_process_resolver`)
- Tested against `v2.0` at commit `2164e61` on Linux x86_64 (Ubuntu 24.04, Clang 18.1.3)

## Change

```diff
 void SignalHandler(int) {
   SPDLOG_INFO("Ctl+C");
   running = false;
-  exit(0);
 }
```

### DIFF
--- a/shell/main.cc
+++ b/shell/main.cc
@@ -39,7 +39,6 @@ std::unique_ptr<Logging> gLogger;
 void SignalHandler(int /* signal */) {
   SPDLOG_INFO("Ctl+C");
   running = false;
-  exit(0);
 }
 
 /**


jwinarske:
@aki1770-del 
Incoming merge from this branch:
https://github.com/toyota-connected/ivi-homescreen/tree/jw/reliability-fixes

aki1770-del:
> @aki1770-del Incoming merge from this branch: https://github.com/toyota-connected/ivi-homescreen/tree/jw/reliability-fixes

Thank you for the heads-up, @jwinarske. I can see that jw/reliability-fixes includes 3d1348ce which addresses the same exit(0) in the SIGINT handler. Happy to close this PR in favour of the incoming merge. Looking forward to contributing to any gaps after jw/reliability-fixes lands.

PR #192, opened 2026-05-23
[scripts] sandboxed Raspberry Pi cross-build

jwinarske:
## Summary

- New `scripts/build_pi.sh` — hermetic aarch64 cross-build for PiOS bookworm/trixie. Toolchain (ARM GNU), sysroot (loopback-mounted PiOS image + qemu-chroot apt), and Flutter engine SDK (from `meta-flutter/flutter-engine` releases) all cached under `$XDG_CACHE_HOME/ivi-homescreen-xc` without touching `/usr`. Defaults to `--backend all`, which builds wayland-egl, wayland-vulkan, drm-kms-egl, and software into per-backend build dirs (drm-kms-egl skipped on bookworm — libdisplay-info 0.1.1 < drm-cxx 0.2.0). Toolchain version tracks PiOS so the bundled glibc stays compatible (12.3.rel1 for bookworm, 15.2.rel1 for trixie — glibc 2.41 reshuffled `pthread_cond_t`, breaking older toolchains).
- `scripts/build_drm_bundle.sh` — define `die()` before its first call (unset `FLUTTER_WORKSPACE` previously printed `die: command not found`), and let `ENGINE_BUNDLE` be set directly so the script can consume the aarch64 engine SDK that `build_pi.sh` caches.
- `shell/backend/drm_kms_egl/drm_backend.cc` — guard the `compositor_` reference in `StopVsyncMonitor()`'s force-clear path behind `#if BUILD_COMPOSITOR` (member is declared only under that guard), and mark the `SPDLOG_DEBUG`-only `latency_ms` `[[maybe_unused]]`. Both were latent in the source but only fire when building with `BUILD_BACKEND_DRM_KMS_EGL=ON` + `BUILD_COMPOSITOR=OFF`, which is the cross-build script's default.

## Test plan

- [x] `scripts/build_pi.sh --pios trixie` builds all four backends (wayland-egl / wayland-vulkan / drm-kms-egl / software) into per-backend dirs; each produces an `ARM aarch64` ELF.
- [x] `scripts/build_pi.sh --pios bookworm` builds three backends (drm-kms-egl auto-skipped); each produces an `ARM aarch64` ELF.
- [x] `scripts/build_pi.sh --prepare-only` fetches toolchain + sysroot + engine SDK without configuring.
- [x] `scripts/build_drm_bundle.sh` with `ENGINE_BUNDLE` set assembles a `.desktop-homescreen/` bundle whose `lib/libflutter_engine.so` is `ARM aarch64` and `data/icudtl.dat` is present.
- [x] Run a bundled `homescreen` on actual Pi 4 / Pi 5 / Pi Zero 2 W hardware (not yet exercised).

### DIFF
--- a/scripts/build_drm_bundle.sh
+++ b/scripts/build_drm_bundle.sh
@@ -29,17 +29,21 @@ set -euo pipefail
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 REPO_DIR="$(dirname "$SCRIPT_DIR")"
 
+die() { echo "error: $*" >&2; exit 1; }
+
 # ── Configurable paths ───────────────────────────────────────────────────
 
-[[ -n "${FLUTTER_WORKSPACE:-}" ]] || die "FLUTTER_WORKSPACE env var must be set"
-ENGINE_BUNDLE="${ENGINE_BUNDLE:-${FLUTTER_WORKSPACE}/flutter-engine/bundle-debug-x64}"
-CONFIG_TEMPLATE="${CONFIG_TEMPLATE:-${FLUTTER_WORKSPACE}/desktop-homescreen/config.toml}"
+# ENGINE_BUNDLE may also be set directly (skips the FLUTTER_WORKSPACE check).
+if [[ -z "${ENGINE_BUNDLE:-}" ]]; then
+    [[ -n "${FLUTTER_WORKSPACE:-}" ]] \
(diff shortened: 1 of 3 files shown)


jwinarske:
cc @kerberjg script has pattern of fetching OS image + generates sysroot for cross compiling.  The script cross-compiles all backend variants.

PR #200, opened 2026-05-30
[drm_kms_egl] wire LayerScene through the non-framed DrmCompositor present path

jwinarske:
## Summary

Lights up `USE_DRM_SCENE` end-to-end on top of the wrapper scaffolding
from PR #199. The OFF build is byte-for-byte identical to `v2.0` (no
`scene_` member, no scene includes, no behavioural change). The ON
build constructs a `drm::scene::LayerScene` in `InitPlaneAllocator`
and routes every non-framed present through it. Framed mode is
unchanged and continues to use the manual primary-BG + overlay path
(LayerScene's allocator can't produce that shape on amdgpu DC).

## Update — primary direct-scanout + per-platform REFLECT_Y gating

Follow-on commits bump drm-cxx to pick up the LayerScene fixes from
jwinarske/drm-cxx#77 and make the scene path actually achieve zero-copy
direct scanout where the hardware allows it, with a clean fallback where
it doesn't. Validated on RPi5 vc4 and amdgpu DC (see Verification).

**drm-cxx#77** — three bugs in the previously-unreached primary-assignment path:
- `Allocator::apply` resolved the CRTC index as the first populated `possible_crtcs` bit instead of the Output's actual CRTC, so on multi-CRTC parts (RPi5 vc4: chosen CRTC at index 2, overlays span `0xe`) it bound a foreign-CRTC plane → EINVAL on every commit.
- writing a fixed-range (`zpos_min == zpos_max`) but non-`IMMUTABLE`-flagged zpos (vc4 primary advertises `[0,0]` without the flag) → EINVAL; now also skipped.
- `remove_layer` left dangling `scene_layer` pointers in the deferred-release ring (`prev_/prev_prev_acquisitions_`) → use-after-free in `release_all` on the next `finalize_frame`.

**Embedder side (this PR):**
- `ProbePrimaryReflectY` — a one-shot `TEST_ONLY` atomic commit that decides per device whether the primary actually accepts `ROTATE_0 | REFLECT_Y` with the backing-store format. The rotation-enum bitmask is unreliable (it false-positives on vc4), so the probe is the authoritative signal.
- `InitDirectOverlay` / `PresentDirectOverlay` — zero-copy direct-overlay path for hardware whose primary can't REFLECT_Y but an overlay can: an opaque background is pinned to the primary (full CRTC coverage) and the backing store is direct-scanned on a REFLECT_Y overlay. Falls back to GL for multi-layer / platform-view frames.
- **Scene gated on REFLECT_Y availability** — when no plane supports REFLECT_Y (amdgpu DC: primary + both overlays advertise ROTATE_0 only), `scene_` is not constructed; the present falls through to the legacy GL-composite path instead of setting REFLECT_Y on every candidate plane, failing every atomic TEST, and dropping to a slow `CompositeCanvas` read-back.

| Platform | Primary REFLECT_Y | Present path | Result |
|---|---|---|---|
| RPi5 vc4 | yes | scene direct-scanout on primary | zero-copy, ~60fps |
| amdgpu DC | no (no plane) | legacy GL-composite (scene skipped) | no EINVAL, right-side-up |
| no-REFLECT_Y primary + REFLECT_Y overlay | n/a | direct-overlay (BG on primary + BS on overlay) | zero-copy |

### `DrmCompositor.h`
- `StoreBaton` grows a `scene_source: std::unique_ptr<GbmBackingStoreLayerSource>`, pre-built in `CreateBackingStore` and consumed by the scene's first `add_layer` for that baton.
- New member `scene_: std::unique_ptr<drm::scene::LayerScene>`.
- New member `scene_layer_batons_: vector<StoreBaton*>` mirroring the scene's live layer set so a per-frame prune can remove layers whose baton fell out of the current `FlutterLayer[]` (drm-cxx's scene exposes no iteration API beyond `find_by_identity_tag`, so the prune list lives embedder-side).
- New private `PresentLayersViaScene(...)` declaration.
- Direct-overlay members (`direct_overlay_mode_`, `direct_overlay_id_`, `direct_overlay_zpos_`, `primary_supports_reflect_y_`) and `InitDirectOverlay` / `ProbePrimaryReflectY` / `PresentDirectOverlay` declarations; `CreateGbmStore` gains a `force_scanout` parameter for stores created before `planes_available_` is set.

### `DrmCompositor.cc`
- `InitPlaneAllocator` constructs the LayerScene after `framed_` detection (skipped in framed mode, and now also skipped when no plane supports REFLECT_Y) from `backend_->device()` + crtc/connector ids + mode. When the primary can't REFLECT_Y but an overlay can, it adopts direct-overlay mode instead.
- `CreateBackingStore` pre-builds the `GbmBackingStoreLayerSource` and stashes it in the baton; `CollectBackingStore` removes the matching scene layer (via `find_by_identity_tag`) before deleting the baton.
- `PresentLayers` dispatches to `PresentDirectOverlay` / `PresentLayersViaScene` under `USE_DRM_SCENE` for the non-framed branch.
- New `PresentLayersViaScene`: walks the `FlutterLayer[]` in z-order, syncs the scene's layer set against this frame's batons (add new, `set_dst_rect_if_changed` on existing, remove stale), runs `scene_->test()` to predict placements, falls through to `PresentViaGlFallback` when any layer would be `Composited` or `Unassigned` (GL composite path is faster than `CompositeCanvas`'s CPU read-back over uncached GBM) or when any FlutterLayer is a platform view (PV→LayerBufferSource plumbing is a follow-on). Otherwise `glFinish`-syncs GPU writes, `scene_->commit`'s with `PAGE_FLIP_EVENT | NONBLOCK`, handles `EACCES` skip-frame and latched fallback, advances first-commit modeset state + `VerifyPipeRunning` + initial vsync baton, and rotates the BS slot pipeline identically to the legacy path. Debug logging emits per-frame `CommitReport` counters plus a separate scene-profile block under `IVI_DRM_PROFILE=1`.
- `SetPaused` mirrors pause edges into `scene_->on_session_paused`.
- `OnResume` calls `scene_->on_session_resumed(backend_->device())` after the existing latch reset; `DrmSession::TakeDevice` already opts in to `preserve_fd_across_resume` so the wrapped `Device`'s fd is the same one across pause/resume cycles.

## Out of scope (deferred)

- **Member cleanup** — `plane_registry_`, `allocator_`, `comp_layer_`, `modeset_`, `plane_mode_set_`, `primary_zpos_` remain because framed-mode reads `plane_registry_` for plane discovery and `primary_zpos_` for overlay zpos placement. Removing them needs framed-mode also migrated to LayerScene — separate PR.
- **Platform-view→LayerBufferSource plumbing** — frames containing a platform view force the GL composite path. PV adapters (Tier-2 wrappers over `ExternalDmaBufSource` / `GstAppsinkSource`) are follow-on work.
- **Explicit fence sync on the scene path** — the legacy direct-scanout path uses `IN_FENCE_FD` when the EGL native-fence-sync extension is present; the scene path falls back to a synchronous `glFinish` before commit. Correctness-equivalent at the cost of one per-frame stall on Tegra; `LayerBufferSource` sync hooks are a follow-on.

## Verification

- [x] `USE_DRM_SCENE=ON` / `OFF` — DRM backend builds clean (clang-tidy + clang-format clean); OFF is identical to `v2.0` (no `scene_` symbols, no `drm::scene` includes).
- [x] **RPi5 vc4** (trixie, kernel 6.12, 1280x1440) — backing store direct-scans on the **primary** with REFLECT_Y: `atomic TEST FAIL: 0`, no `partial-fallback drop`, steady ~60fps (`commit ~0.1ms`, no compositing), 0 `[W]`, right-side-up, no crash over a full run.
- [x] **amdgpu DC** (native x86_64, ICS-homescreen bundle) — no plane supports REFLECT_Y → scene gated off → legacy GL-composite path: 0 EINVAL, 0 `partial-fallback drop`, no crash, right-side-up. Replaces the prior behaviour where the scene set REFLECT_Y unconditionally and fell to a slow `CompositeCanvas` read-back with per-frame EINVAL.

### DIFF
--- a/scripts/build_pi.sh
+++ b/scripts/build_pi.sh
@@ -39,6 +39,11 @@
 #                                   (engine is dlopen'd at runtime — not linked at build)
 #     --plugins-dir <path>          default: ../ivi-homescreen-plugins/plugins
 #     --no-plugins                  build homescreen only
+#     --with-scene                  drm-kms-egl only: enable the plane
+#                                   compositor + drm-cxx LayerScene present
+#                                   path (BUILD_COMPOSITOR + USE_DRM_SCENE).
+#                                   Default off = direct DRM/KMS + GBM, no
+#                                   compositor.
 #     --jobs <N>                    default: nproc
 #     --clean                       wipe build dir before configure
 #     --prepare-only                fetch/extract toolchain, sysroot, engine, then exit
@@ -146,6 +151,7 @@ ENGINE_URL=""
 # actual CMake entry; the repo root has no CMakeLists.txt.
 PLUGINS_DIR="${REPO_DIR%/*}/ivi-homescreen-plugins/plugins"
 NO_PLUGINS=0
+WITH_SCENE=0
 JOBS="$(nproc 2>/dev/null || echo 4)"
 CLEAN=0
 PREPARE_ONLY=0
@@ -236,6 +242,7 @@ while [[ $# -gt 0 ]]; do
         --engine-url)         ENGINE_URL="$2"; shift 2 ;;
         --plugins-dir)        PLUGINS_DIR="$2"; shift 2 ;;
         --no-plugins)         NO_PLUGINS=1; shift ;;
+        --with-scene)         WITH_SCENE=1; shift ;;
         --jobs)               JOBS="$2"; shift 2 ;;
         --clean)              CLEAN=1; shift ;;
         --prepare-only)       PREPARE_ONLY=1; shift ;;
@@ -922,7 +929,13 @@ phase4_build() {
                 -DBUILD_BACKEND_WAYLAND_EGL=OFF
                 -DBUILD_BACKEND_WAYLAND_VULKAN=OFF
                 -DBUILD_BACKEND_DRM_KMS_EGL=ON
-                -DBUILD_BACKEND_SOFTWARE=OFF) ;;
+                -DBUILD_BACKEND_SOFTWARE=OFF)
+            # --with-scene lights up the plane compositor + drm-cxx LayerScene
+            # present path (direct-scanout with GL fallback) instead of the
+            # default direct DRM/KMS + GBM path. Only meaningful for drm-kms-egl.
+            if [[ "$WITH_SCENE" -eq 1 ]]; then
+                cmake_args+=(-DBUILD_COMPOSITOR=ON -DUSE_DRM_SCENE=ON)
+            fi ;;
         software)
             cmake_args+=(
                 -DBUILD_BACKEND_WAYLAND_EGL=OFF
--- a/shell/backend/drm_kms_egl/drm_backend.cc
+++ b/shell/backend/drm_kms_egl/drm_backend.cc
@@ -1426,7 +1426,17 @@ void DrmBackend::UnifiedPageFlipHandler(int /*fd*/,
     spdlog::info("[DrmBackend] flip #{} handled (asio monitor)", n);
   }
 #if BUILD_COMPOSITOR
-  if (self->compositor_ && self->compositor_->planes_active()) {
+  // Dispatch the completion to whichever subsystem ARMED this flip, identified
+  // by its pending flag — NOT by planes_active(). planes_active() reflects who
+  // owns scanout in steady state, but when the compositor is active and a frame
+  // falls back to backend_->Present()'s legacy page flip (e.g. a backing-store
+  // layer needs composition), the completion must clear
+  // DrmBackend::flip_pending_. Routing by planes_active() cleared the
+  // compositor's (unset) flag instead, leaving DrmBackend::flip_pending_ stuck
+  // and deadlocking the next WaitForPendingFlip(). At most one flip is in
+  // flight per CRTC, so at most one flag is set; the compositor's pending flag
+  // therefore identifies its own flips.
+  if (self->compositor_ && self->compositor_->IsFlipPending()) {
     self->compositor_->OnFlipComplete();
   } else {
     self->OnLegacyFlipComplete();
--- a/shell/backend/drm_kms_egl/drm_compositor.cc
+++ b/shell/backend/drm_kms_egl/drm_compositor.cc
@@ -20,6 +20,7 @@
 #include <poll.h>
 #include <unistd.h>
 
+#include <algorithm>
 #include <array>
 #include <cerrno>
 #include <chrono>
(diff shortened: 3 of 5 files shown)


jwinarske:
## Validation — rebased onto current `v2.0` + VKMS smoke

Rebased `feat/drm-scene-compositor` onto current `v2.0` (was 16 behind; the only overlap was a 9-line `main.cc` hunk from the idle-wakeup commit, auto-merged with no conflict). Diff unchanged at +473/-12 across the same 3 files. Force-pushed.

### Build / static (clang-21, Fedora)
- ✅ `USE_DRM_SCENE=OFF` (default, DRM-EGL + `BUILD_COMPOSITOR=ON`) — clean, no warnings.
- ✅ `USE_DRM_SCENE=ON` — clean, no warnings.
- ✅ `clang-format` clean on all three changed files.

### Runtime — `test/drm_kms_vkms.sh` + direct runs on `vkms` (software/llvmpipe, `IVI_DRM_PROFILE=1`)

Backend brings up cleanly on the virtual KMS node: DRM master, atomic modeset (1024x768@60), EGL config selection, `DrmCursor` ready, engine + app run, backing stores create/collect, **0 `[E]`/`[C]`, no crashes, backing-store pool peak=1 (no leak)**.

What the **`USE_DRM_SCENE=ON`** path exercised depends on whether the probe enables the plane compositor:

| vkms config | Result |
|---|---|
| default (`enable_overlay=0`) | `use_plane_compositor=false` ("disabled by probe; GL path") — `PresentLayers` bails at the `!planes_available_` guard, scene path not reached. Confirms **no-regression**: ON build behaves exactly like the GL path. |
| `enable_overlay=1 enable_plane_pipeline=1` | `compositor=planes`, `LayerScene constructed`, `plane allocator active`, **`PresentLayersViaScene` reached** — `scene_->test()`, `add_layer`/`find_by_identity_tag`, and the fallback decision all run and **gracefully route to `PresentViaGlFallback`** (0 errors) across multiple frames. |

### What VKMS could **not** validate — `scene_->commit()` direct scanout

With overlays on, the single Flutter UI layer is bucketed as **"needs composition"** and dropped (`scene::LayerScene: layer 1 needs composition but its source map() failed (Function not implemented)`), so it never reaches the `scene_->commit()` success path. Two virtual-stack causes, **neither a PR defect**:
1. vkms's primary plane advertises only `XRGB8888` while the backing store is `ARGB8888` — the format mismatch forces composition instead of direct assignment. On hardware whose primary accepts the BS format, the layer assigns directly.
2. The composition path's `gbm_bo_map()` returns `ENOSYS` on llvmpipe, so even the composite-then-scanout route can't run under software rendering.

The PR's fallback logic handles both correctly — graceful GL fallback, no error, no latch-thrash.

### Net

Everything reachable headlessly is green: rebase, both builds, format, no-regression, scene construction, the `PresentLayersViaScene` decision/sync machinery, and graceful fallback. The remaining **`[ ] Real hardware`** checklist row (the `scene_->commit()` direct-scanout success path) genuinely needs a real DRM device whose primary plane accepts the backing-store format — it cannot be exercised on vkms+llvmpipe. On this box amdgpu (`card1`) is held by the Wayland session, so that row needs a free VT / second card to close out before merge.

PR #211, opened 2026-06-02
[drm_kms_egl] fix cold-start stall: deliver the parked first vsync baton

### LINKED ISSUE #210: DRM/KMS: intermittent cold-start stall — black screen with only the HW cursor (Flutter never presents first frame)
## Summary

The DRM/KMS-EGL backend intermittently fails to start: on a cold start the
screen comes up **black with only the hardware mouse cursor visible**, and the
app never produces any frames. **Restarting the app resolves it** — it's a
per-cold-start race, hit roughly 30–40% of the time.

Observed repeatedly while testing the scene compositor on **vc4 (Raspberry Pi 5
and Raspberry Pi 4, Trixie)**. It is **not** specific to the scene path — see
"Not caused by the present path" below.

## Symptom

- Black screen, only the HW cursor drawn (cursor is movable / visible).
- No Flutter content ever appears.
- Killing and re-launching the app almost always succeeds on the next try.

## Log signature

The log is short and stops progressing right after the engine bootstraps. The
**tell is what's missing**: there is no `[DrmCompositor] N planes available for
CRTC …` line (the compositor's lazy init), no `PresentLayers`, and no `FPS:` —
i.e. the embedder compositor is never entered, so **Flutter never presents its
first frame**.

Representative tail of a stalled run (vc4 RPi5, card1, 1280x1440@60):

```
[23:10:10.062] [I] [DrmBackend] opened /dev/dri/card1 via libseat (fd=5)
[23:10:10.114] [I] [DrmBackend] connector=33 crtc=92 mode=1280x1440@60Hz
[23:10:10.115] [I] [DrmBackend] driver='vc4' compositor=planes modeset=atomic ... overlay-planes=yes explicit-sync=yes
[23:10:10.191] [I] [DrmCursor] ready (sprite=24px buffer=64px path=atomic-cursor plane_id=655 letterbox=0,0)
[23:10:10.193] [I] (0) Loading AOT: .../lib/libapp.so
[23:10:10.198] [I] [DrmBackend] flip monitor armed on fd=5
[23:10:10.198] [I] [DrmBackend] SendWindowMetrics 1280x1440 result=0
[23:10:10.208] [I] flutter: bootstrap start...
[23:10:10.212] [I] flutter: Saving...
[23:10:10.214] [I] flutter: Navigate to: /home
[23:10:10.220] [I] [DrmSeat] started
[23:10:12.213] [I] flutter: Saving...
                          <-- ~20 s of silence; no compositor init, no present -->
[23:10:32.057] [E] [DrmSession] seat fd error (revents=0x11); exiting   (teardown from the test harness timeout)
[libseat error] [libseat/backend/seatd.c:122] Could not flush connection: Broken pipe
homescreen exit 124
```

A healthy start instead continues past `[DrmSeat] started` with
`[DrmCompositor] N planes available for CRTC …`, `PresentLayers entered …`, and
`[DrmBackend] FPS: …`.

The HW cursor renders because it is programmed on its own atomic-cursor plane at
`DrmBackend` init, independent of the Flutter present path — so the CRTC/pipe is
live; only Flutter content never arrives.

## Not caused by the present path

The **same binary** stalls on one cold start and renders fine on the next, and
the compositor present code (`CreateBackingStore` / `PresentLayers` /
`EnsureGlCapsProbed`) never executes in the stalled case. So this is independent
of the scene-compositor / direct-overlay work — it is a first-frame bootstrap
race in the backend startup, surfaced by repeated cold starts.

## Suspected cause (not yet root-caused)

The initial vsync-baton hand-off between `FlutterEngineRun` and the backend's
first `OnVsync` kick (`vsync_baton_` / `PostOnVsync` / the first-commit
baton-drain on the Present* paths). If the engine's first frame request and the
backend's initial kick race, the frame pacer can wedge before the first present,
leaving Flutter with no outstanding vsync and the backend with no frame.

## Workaround

Restart the app; the next cold start almost always proceeds.

## Environment

- Backend: `BUILD_BACKEND_DRM_KMS_EGL=ON`, `BUILD_COMPOSITOR=ON`, `USE_DRM_SCENE=ON`
- HW: vc4 — reproduced on Raspberry Pi 5 and Raspberry Pi 4, Raspberry Pi OS Trixie (aarch64)

jwinarske:
## Summary

Fixes the intermittent DRM cold-start stall (#210): on ~30–40% of cold starts the
screen comes up black with only the hardware cursor and the app never renders;
a restart clears it.

## Root cause

`FlutterView` runs the engine (`flutter_view.cc:405`) **before** wiring the
platform task runner (`SetPlatformTaskRunner`, `flutter_view.cc:419`). If the
engine's first vsync request lands in that window:

1. `SetVsyncBaton` sees the pipeline idle (`flip_in_flight == false`) and takes
   the inline-kick branch: `vsync_baton_.exchange(0)` pulls the baton out, then
   `PostOnVsync` is called.
2. `PostOnVsync` finds `platform_task_runner_ == nullptr` and **drops the
   baton** — and it was already exchanged out of `vsync_baton_`, so it's lost.
3. `SetPlatformTaskRunner` then wires the runner and arms the flip monitor but
   does **nothing** to recover the dropped baton (the "kick latch" its own
   comment referenced never existed).

Flutter waits forever for the first `OnVsync`, never schedules frame 1, so the
rasterizer never calls `CreateBackingStore`/`PresentLayers` and the compositor's
lazy init never runs — black screen (the cursor shows because it's on its own
atomic-cursor plane). Confirmed live with temporary instrumentation:

```
first SetVsyncBaton: baton=547474122864 flip_in_flight=false runner_wired=false
PostOnVsync DROP: platform runner not wired (runner=0x0), baton=547474122864 LOST
flip monitor armed on fd=5      (~210 µs too late)
```

## Fix

- **`SetVsyncBaton`** — only take the baton out of `vsync_baton_` and kick it
  when the platform task runner is wired and can deliver it; otherwise leave it
  *parked* in `vsync_baton_` for the kick latch. (Gating before the exchange,
  rather than exchange-then-restore, keeps the hand-off to the latch race-free —
  the transient empty state of exchange-then-restore can race the latch and
  strand the baton.)
- **`SetPlatformTaskRunner`** — add the kick latch: after storing the runner and
  starting the flip monitor, deliver any parked baton. The runner store
  (release) before the baton `exchange` (acq-rel) makes the hand-off deliver the
  first baton exactly once, by whichever path first observes the runner wired.
- **`PostOnVsync`** — drop-path comment updated (callers now park instead of
  stranding).

## Verification

- **RPi4 vc4** (Trixie, 1920x1080@60): a 20-iteration cold-start loop now
  presents **20/20** (`present=20 stall=0`), each rendering via the scene
  direct-scanout path. Baseline was ~30–40% stalls, so 0/20 puts the odds of a
  fluke at ~0.08%.
- **amdgpu DC**: unaffected — still presents (gated to the GL-composite path),
  no crash.
- clang-tidy / clang-format clean on the change.

Closes #210.

### DIFF
--- a/shell/backend/drm_kms_egl/drm_backend.cc
+++ b/shell/backend/drm_kms_egl/drm_backend.cc
@@ -1267,10 +1267,11 @@ void DrmBackend::PostOnVsync(FLUTTER_API_SYMBOL(FlutterEngine) engine,
   }
   auto* runner = platform_task_runner_.load(std::memory_order_acquire);
   if (runner == nullptr || runner->GetStrandContext() == nullptr) {
-    // Plumbing not in place yet — drop. Happens between vsync_callback
-    // firing during engine bring-up and FlutterView::SetPlatformTaskRunner
-    // landing. The kick latch ensures Flutter still gets a baton from
-    // the next-frame path once the runner is wired.
+    // Runner not wired yet (the engine's first vsync request raced
+    // SetPlatformTaskRunner during bring-up). Callers leave the baton
+    // parked in vsync_baton_ in this case rather than exchanging it out,
+    // so SetPlatformTaskRunner's kick latch delivers it once the runner
+    // is wired (#210).
     return;
   }
   const uint64_t now = LibFlutterEngine->GetCurrentTime();
@@ -1308,9 +1309,20 @@ void DrmBackend::SetVsyncBaton(FLUTTER_API_SYMBOL(FlutterEngine) engine,
     return;
   }
 
-  // Exchange to take the baton back. The asio handler may have raced
-  // us and already consumed it; exchange returning 0 means it's
-  // already on its way.
+  // Idle pipeline — kick the baton inline. Only pull it out of
+  // vsync_baton_ if the platform task runner is actually wired and can
+  // deliver it; otherwise leave it parked so SetPlatformTaskRunner's kick
+  // latch delivers it once wired. At cold start the engine's first vsync
+  // request can land before SetPlatformTaskRunner; exchanging the baton
+  // out only to drop it in PostOnVsync stranded Flutter's first frame
+  // (#210) — frame 1 was never produced (black screen + only the HW
+  // cursor). Gating here (rather than exchange-then-restore) keeps the
+  // hand-off to the kick latch race-free: the baton is taken exactly once,
+  // by whichever of this path / the latch first sees the runner wired.
+  auto* runner = platform_task_runner_.load(std::memory_order_acquire);
+  if (runner == nullptr || runner->GetStrandContext() == nullptr) {
+    return;  // parked in vsync_baton_; SetPlatformTaskRunner drains it
+  }
   if (const intptr_t mine = vsync_baton_.exchange(0, std::memory_order_acq_rel);
       mine != 0) {
     PostOnVsync(engine, mine);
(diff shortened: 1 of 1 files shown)


jwinarske:
Follow-up 40/40 test ran, no sign of a stall.

PR #232, opened 2026-06-07
[drm_kms_egl] opt-in raster-thread page-flip drain (IVI_DRM_RASTER_DRAIN)

jwinarske:
## Problem

`WaitForPendingFlip()` (rasterizer thread) relies on the asio flip monitor on the **platform task runner thread** to drain `PAGE_FLIP_EVENT`s and clear `flip_pending_`. When that thread is starved — CPU cpufreq governor downclock (`ondemand`), or a long Dart GC pause — the kernel fires the event but nobody reads it before the 100ms deadline. The rasterizer then logs `PAGE_FLIP_EVENT likely lost` and proceeds while the flip is still in flight, so the next `drmModePageFlip` can `EBUSY` and leak the pending BO.

Root-caused on i.MX8MM: the event is **not** lost — it's sitting readable on the fd, just undrained because the monitor thread didn't get scheduled.

## Fix (opt-in, default off)

`IVI_DRM_RASTER_DRAIN=1` lets the rasterizer drain the fd itself via `DrainFlipEvents()`:
- `poll(0)` + `drmHandleEvent()` **serialized with the asio monitor by `drm_event_mutex_`**. The poll and read run under the lock so the read can never block — only one thread consumes a given event; the loser's `poll(0)` then shows nothing readable.
- The asio monitor and the teardown drain in `StopVsyncMonitor()` route through the same helper, so there's a single guarded drain path.

**Default off → no behavior change on validated platforms.** In the healthy baton-driven flow `flip_pending_` is already false when `Present()` runs, so `WaitForPendingFlip()` returns before touching the lock — zero overhead. The drain only ever runs in the starved case.

## Self-diagnostic

When the flag is **off**, a read-only `poll(0)` probe at the deadline distinguishes a *starved monitor* (event readable but undrained → the flag would help) from a *genuinely late flip* (no event yet → it wouldn't), and logs a one-time, actionable recommendation to set `IVI_DRM_RASTER_DRAIN=1`. `poll(0)` doesn't `read()`, so it neither consumes the event nor races the monitor.

## Reliability / security / performance

- **Security:** none — no new input, the DRM fd is already trusted.
- **Performance:** none on the healthy path (returns before lock/poll); in the starved case it's *lower* latency than the sleep-wait. Single mutex → no lock-ordering deadlock. Invariant documented: nothing reachable from `UnifiedPageFlipHandler` may take `drm_event_mutex_` (the handler runs under it inside `drmHandleEvent`).
- **Reliability:** removes the EBUSY/leaked-BO failure mode behind the give-up.

## Testing

Validated on i.MX8MM (Nitrogen8MM, imx-drm) under the `ondemand` governor:
- flag **off** → `WaitForPendingFlip` warning + the one-time `Set IVI_DRM_RASTER_DRAIN=1` recommendation.
- flag **on** → **0** warnings, renders cleanly, no abort/deadlock.

⚠️ Touches the shared flip path used by the compositor/scene present on amdgpu / rk3588 / msm / vc4, which I could only build- but not run-test here. Recommend a quick sanity check on an amdgpu or RPi target before flipping the default on in a later change.

### DIFF
--- a/shell/backend/drm_kms_egl/drm_backend.cc
+++ b/shell/backend/drm_kms_egl/drm_backend.cc
@@ -1594,10 +1594,7 @@ void DrmBackend::StopVsyncMonitor() {
       pollfd pfd{drm_dev_->fd(), POLLIN, 0};
       const int rc = ::poll(&pfd, 1, kPollSliceMs);
       if (rc > 0 && (pfd.revents & POLLIN)) {
-        drmEventContext ctx{};
-        ctx.version = 2;
-        ctx.page_flip_handler = &DrmBackend::UnifiedPageFlipHandler;
-        drmHandleEvent(drm_dev_->fd(), &ctx);
+        DrainFlipEvents();
       } else if (rc < 0 && errno != EINTR) {
         break;
       }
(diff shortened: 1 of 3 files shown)


jwinarske:
## amdgpu sanity check ✅

Ran the branch natively on a host **amdgpu** GPU (Radeon `1002:13C0`, `/dev/dri/card1`, atomic modeset, `compositor=planes`, overlay planes, async-flip) with a JIT bundle, default off and flag on:

| run | result |
|---|---|
| flag **off** (default) | 60 Hz, 84% in the 60Hz bucket, `present_failures=0`, renders clean |
| flag **on** (`IVI_DRM_RASTER_DRAIN=1`) | 98% in the 60Hz bucket, `present_failures=0`, **0** errors, no deadlock |

**No regression on the shared flip path.** amdgpu drives scanout through `compositor=planes`, whose flip-completion events are drained by the asio monitor — i.e. the `ArmFlipRead()` → `DrainFlipEvents()` (mutex + `poll(0)`) change this PR makes. That path runs clean at 60 Hz with `present_failures=0`, confirming the mutex-guarded drain doesn't deadlock or drop events on amdgpu.

Note: the `IVI_DRM_RASTER_DRAIN` flag itself is a **no-op on the compositor (planes) path** — that path has its own `DrmCompositor::WaitForPendingFlip`, separate from the `DrmBackend::WaitForPendingFlip` this PR touches (used by the legacy-GL path, e.g. imx-drm `compositor=gl`). So the flag is safe-and-inert on amdgpu/rk3588/msm; extending the same drain to `DrmCompositor` would be a follow-up if those platforms ever hit the starvation.

(The pre-existing `[DrmCompositor]` warning seen once was during a ~1.6 s startup stall, and the `exit=137` is the known amdgpu SIGKILL-on-teardown — both unrelated to this change.)

jwinarske:
Follow-up pushed (`edeba504`): extended `IVI_DRM_RASTER_DRAIN` to **`DrmCompositor::WaitForPendingFlip`** (the `compositor=planes` path used by amdgpu/rk3588/msm), so the flag now covers both the legacy-GL and plane-compositor flip waits. It reuses the same mutex-guarded `DrainFlipEvents()` (DrmCompositor is already a friend of DrmBackend); `OnFlipComplete` is safe from the rasterizer thread because every drain path holds `drm_event_mutex_`, `flip_pending_` is atomic, and the slot-release pipeline has its own lock. Validated on amdgpu (`compositor=planes`): flag on → 60Hz, `present_failures=0`, no deadlock, no stall warnings. Default still off.

jwinarske:
## RPi5 (V3D/vc4) sanity ✅ — exercises the compositor drain under real starvation

Ran on a Raspberry Pi 5 (Debian trixie, glibc 2.41; **`driver=vc4`, `compositor=planes`**, atomic, overlay planes, async-flip), `/dev/dri/card1`, under the Pi's default **`ondemand`** governor — which naturally reproduced the platform-thread starvation:

| run | `[DrmCompositor]` stall warns | "readable but undrained" recommendation | present_failures | errors/deadlock |
|---|---|---|---|---|
| flag **off** (default) | 2 | 1 | 0 | none |
| flag **on** (`IVI_DRM_RASTER_DRAIN=1`) | **0** | 0 | 0 | **none** |

So on a second `compositor=planes` platform (and a different GPU/SoC), the new `DrmCompositor` drain path:
- **with the flag off**, the diagnostic correctly detected the readable-but-undrained event and recommended the flag (real `ondemand` starvation, not synthetic);
- **with the flag on**, the stall warnings are eliminated (max_interval 1.28s → 0.68s), no deadlock, `present_failures=0`.

Combined coverage now: GL path (`DrmBackend`) validated on imx-drm; planes path (`DrmCompositor`) validated on amdgpu (no-regression) and **RPi5 (starvation reproduced + fixed)**.

jwinarske:
## rk3588 (rockchip VOP2) sanity ✅ no regression / no deadlock — flag benefit is partial here

Ran on a FriendlyElec NanoPC-T6 (rk3588, Debian trixie, **Mali blob**; `driver=rockchip`, **`compositor=planes`** direct-scanout, atomic, overlay planes), `/dev/dri/card0`, both `conservative` (default) and `performance` governors, flag off/on — 4 runs.

**Reliable flip-path signals (all 4 runs):** `present_failures=0`, no deadlock, no EBUSY / no GL fallback (direct-scanout stayed healthy). So **flag off = no regression, flag on = no deadlock / no new errors.**

**Flag effect is partial here (unlike RPi5):**
- Flag off → the `[DrmCompositor]` "readable but undrained" recommendation fires, so *some* stalls are the drain-fixable starvation.
- Flag on → stall warnings drop (`2 → 1`) but not to zero: rk3588 also has occasional **genuinely-late flips** (PAGE_FLIP_EVENT not readable at the 100 ms deadline), which the raster drain can't fix because the event truly isn't there yet.
- rk3588 is **render-bound** on this workload (Mali blob rendering Wonderous sits ~50–100 ms/frame even under `performance`), so the cadence buckets are dominated by GPU time and swing with on-screen content — not a clean fps A/B.

The lone `[E]` seen was `[DrmSession] seat fd error (revents=0x20); exiting` at teardown (SIGTERM → POLLERR on the seat fd), present regardless of the flag.

**Takeaway for default-on:** the flag is safe (no regression / no deadlock) on a third platform and second GPU vendor, but its *benefit* is platform-dependent — big on RPi5/imx-drm, partial on rk3588. Reinforces keeping it opt-in until the matrix + pause/resume pass is complete.

### Coverage so far
| Path | Platform | Regression (off) | Deadlock (on) | Flag fixes stalls |
|---|---|---|---|---|
| `compositor=gl` (`DrmBackend`) | imx-drm | none | none | yes |
| `compositor=planes` (`DrmCompositor`) | amdgpu | none | none | n/a (no starvation seen) |
| `compositor=planes` (`DrmCompositor`) | RPi5 vc4 | none | none | yes (2→0) |
| `compositor=planes` (`DrmCompositor`) | rk3588 VOP2 | none | none | partial (2→1) |

jwinarske:
## msm (Qualcomm Adreno / DPU) sanity ✅ — flag clears stalls (2→0)

Ran on an Arduino Uno Q (Qualcomm, Debian trixie; `driver=msm`, **`compositor=planes`**, atomic, overlay planes, **`async-flip=no`**), `/dev/dri/card0`, under the default `schedutil` governor (reused the trixie aarch64 binary; DRM master via the direct-open fallback since card0 was free + user in the `video` group — no seatd needed).

| run | `[DrmCompositor]` stall warns | recommendation | present_failures | errors/deadlock |
|---|---|---|---|---|
| flag **off** | 2 | 1 | 0 | none |
| flag **on** (`IVI_DRM_RASTER_DRAIN=1`) | **0** | 0 | 0 | **none** |

Under `schedutil` the platform-thread monitor was starved → flag-off fired the recommendation; flag-on **eliminated the stalls entirely** (like RPi5), no deadlock, `present_failures=0`. Fourth platform / third GPU vendor (Qualcomm), and the first with `async-flip=no` — the drain works there too.

### Coverage so far
| Path | Platform | GPU vendor | Regression (off) | Deadlock (on) | Flag clears stalls |
|---|---|---|---|---|---|
| `compositor=gl` (`DrmBackend`) | imx-drm | Vivante | none | none | yes |
| `compositor=planes` (`DrmCompositor`) | amdgpu | AMD | none | none | n/a (no starvation seen) |
| `compositor=planes` (`DrmCompositor`) | RPi5 vc4 | Broadcom/V3D | none | none | yes (2→0) |
| `compositor=planes` (`DrmCompositor`) | rk3588 VOP2 | Arm Mali (blob) | none | none | partial (2→1) |
| `compositor=planes` (`DrmCompositor`) | Uno Q msm | Qualcomm Adreno | none | none | yes (2→0) |

Remaining before default-on: **nvidia-drm** (the genuinely-lost-event case — confirm no hang) and a **pause/resume + teardown** pass with the flag on.

jwinarske:
## nvidia-drm sanity ✅ — the lossy-event case, no hang (confirmed empirically)

Ran on a Jetson Orin Nano (L4T R36/JetPack 6). Notably the DRM driver name reports as **`driver='nvidia-drm'`** (platform dev is `nv_platform`, but `drmGetVersion` → `nvidia-drm`), so this *does* hit the code's nvidia-drm quirk path — the exact "PAGE_FLIP_EVENT can be lost" driver. GBM works via JetPack 6's `tegra_gbm` backend; `compositor=planes`, 2560x1440, `schedutil`.

| run | stall warns | recommendation | present_failures | errors | hang/deadlock | render |
|---|---|---|---|---|---|---|
| flag **off** | 2 | 1 | 0 | 0 | no | 89.6% @60Hz |
| flag **on** | 1 | 0 | 0 | 0 | **no** | 91.2% @60Hz, kept rendering to end |

**The key nvidia-drm result:** flag-on did **not hang**. The residual `2→1` stall is precisely a *genuinely-lost* event (nvidia-drm's documented behavior) that the drain can't recover — and it **falls through the bounded 100 ms deadline and proceeds**, exactly the no-hang-by-construction property (the `poll(0)`-gated read never blocks). Empirically confirmed on the actual lossy driver, `present_failures=0` throughout.

### Bonus: raster-drain enables a self-committing cursor on nvidia-drm
nvidia-drm defaults to *staging* the HW cursor into the compositor commit because a self-committing cursor's blocking commit otherwise starves the PAGE_FLIP_EVENT — but that's the exact problem the raster-drain fixes. With `--drm-stage-cursor no` + `IVI_DRM_RASTER_DRAIN=1`: HW cursor plane self-commits (decoupled, smooth) **and** frames stayed smooth at **94% @60Hz, present_failures=0, no deadlock**. So the drain removes the constraint that forced cursor staging on nvidia-drm.

### Final coverage
| Path | Platform | GPU | Regression (off) | Hang/deadlock (on) | Flag clears stalls |
|---|---|---|---|---|---|
| `compositor=gl` | imx-drm | Vivante | none | none | yes |
| `compositor=planes` | amdgpu | AMD | none | none | n/a (no starvation) |
| `compositor=planes` | RPi5 vc4 | Broadcom | none | none | yes (2→0) |
| `compositor=planes` | rk3588 | Arm Mali | none | none | partial (2→1) |
| `compositor=planes` | Uno Q | Qualcomm | none | none | yes (2→0) |
| `compositor=planes` | Jetson | **nvidia-drm** | none | **none** | partial (2→1, residual = lost events, no hang) |

5 platforms / 5 GPU vendors, no regression / no deadlock anywhere; nvidia-drm lossy-event case confirmed safe.

PR #233, opened 2026-06-08
[drm_kms_egl] enable IVI_DRM_RASTER_DRAIN by default (+ nvidia-drm cursor self-commit)

jwinarske:
## Summary

Flip the raster-thread page-flip drain (introduced opt-in in #232) to **on by default**. Not setting the env var now enables it; `IVI_DRM_RASTER_DRAIN=0` is the escape hatch.

```
raster_drain = (env == nullptr) || (env != "0")   // unset → on; "0" → off
```

Applied to both `DrmBackend::WaitForPendingFlip` (GL path) and `DrmCompositor::WaitForPendingFlip` (plane-compositor path). Also drops the now-moot "set `IVI_DRM_RASTER_DRAIN=1`" recommendation diagnostic — it only ran on the flag-off path, which is no longer the default.

## Why now

#232 shipped opt-in to de-risk the new cross-thread fd draining. It's since been validated across **6 platforms / 5 GPU vendors**, and crucially **every** platform exhibited the drain-fixable starvation under its *default* governor with the flag off — so opt-in left the fix unused exactly where it was needed.

| Path | Platform | GPU | Regression (off) | Hang/deadlock (on) | Pause/resume | Teardown vs off |
|---|---|---|---|---|---|---|
| GL (`DrmBackend`) | imx-drm | Vivante | none | none | clean (3/3) | identical (pre-existing rc137) |
| planes (`DrmCompositor`) | amdgpu | AMD | none | none | — | identical (pre-existing rc137) |
| planes | RPi5 vc4 | Broadcom | none | none | — | clean (rc124) |
| planes | rk3588 | Arm Mali | none | none | clean (3/3) | clean (rc124) |
| planes | Uno Q | Qualcomm | none | none | — | clean (rc124) |
| planes | Jetson | **nvidia-drm** | none | **none (no hang)** | — | clean |

- **No regression** with it off; **no deadlock/hang** with it on, including the nvidia-drm lossy-event case (the genuinely-lost event safely falls through the bounded 100 ms deadline — confirmed empirically).
- **Pause/resume** validated on both the GL path (nitrogen8mm) and the plane-compositor path (rk3588) across real VT-switch session-revoke/restore cycles — paused==resumed, rendering resumes, 0 errors.
- **Teardown** is unchanged: exit codes identical flag-off vs on everywhere (the amdgpu/imx-drm SIGKILL is pre-existing slow teardown, not this change).
- **Healthy path has zero overhead** regardless: `WaitForPendingFlip` returns before touching the drain when the flip already completed.

## Rollback

`IVI_DRM_RASTER_DRAIN=0` disables it per-process. The escape hatch stays for one release; once it's proven in the field the flag can be removed and the drain made unconditional.

### DIFF
--- a/shell/backend/drm_kms_egl/drm_backend.cc
+++ b/shell/backend/drm_kms_egl/drm_backend.cc
@@ -1657,16 +1657,17 @@ void DrmBackend::DrainFlipEvents() const {
 
 bool DrmBackend::WaitForPendingFlip() const {
   using namespace std::chrono_literals;
-  // Opt-in (default off). The asio flip monitor normally drains
-  // PAGE_FLIP_EVENTs on the platform task runner thread; when that thread is
-  // starved (CPU governor downclock, Dart GC), the rasterizer here would sleep
-  // out the whole 100ms deadline even though the event already arrived. With
-  // IVI_DRM_RASTER_DRAIN=1 the rasterizer drains the fd itself instead. Off by
-  // default so behavior on validated platforms is byte-for-byte unchanged; the
-  // diagnostic below reports when turning it on would help.
+  // The asio flip monitor normally drains PAGE_FLIP_EVENTs on the platform task
+  // runner thread; when that thread is starved (CPU governor downclock, Dart
+  // GC) the rasterizer would otherwise sleep out the whole 100ms deadline even
+  // though the event already arrived. So the rasterizer drains the fd itself
+  // (DrainFlipEvents, serialized with the monitor by drm_event_mutex_). On by
+  // default — validated across 6 platforms / 5 GPU vendors (no regression, no
+  // deadlock, pause/resume + teardown clean). Set IVI_DRM_RASTER_DRAIN=0 to opt
+  // out (escape hatch).
   static const bool raster_drain = []() {
     const char* env = std::getenv("IVI_DRM_RASTER_DRAIN");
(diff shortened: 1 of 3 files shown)


jwinarske:
## Added: nvidia-drm cursor self-commits when the raster drain is on (`d176500b`)

Follow-up folded in since it's a direct consequence of default-on. nvidia-drm defaulted to *staging* the HW cursor (`stage_cursor=auto`) because a self-committing cursor's blocking commit starves the PAGE_FLIP_EVENT there — which is exactly the starvation the raster drain now removes. So the `auto` heuristic gains one term:

```cpp
stage_cursor = (cfg == kYes) ||
               (cfg == kAuto && driver == "nvidia-drm" && !RasterDrainEnabled());
```

- Drain **on** (default) + nvidia-drm + `auto` → **self-commit** (smooth, frame-decoupled cursor — this is the `--drm-stage-cursor no` behavior I verified on the Jetson, now the default).
- Drain **off** (`IVI_DRM_RASTER_DRAIN=0`) → `auto` reverts to staging on nvidia-drm (the old safe behavior).
- `kYes`/`kNo` still force the choice; non-nvidia-drm unchanged (already self-commits).

Also factored the drain decision into `DrmBackend::RasterDrainEnabled()` so the two `WaitForPendingFlip` paths and this cursor heuristic share one source of truth (no drift). Build + clang-tidy + clang-format clean.

PR #292, opened 2026-07-09
shared: ihs_shared logging, tracing, and config surfaces

jwinarske:
Builds on the `ihs_shared` scaffold (#291, merged) by filling in three of the capability sub-tables the scaffold reserved: **logging**, **tracing**, and **config**. Each is a C ABI wired into `ihs_get_api`, verified with an external C11 consumer. Three reviewable commits.

## Logging (relocation behind the C ABI)
Moves the DLT bridge out of the shell into `libihs_shared.so`, reached only through the `ihs_dlt_*` C ABI, so the shell and any FFI plugin share one copy of the bridge / ring registry / worker.

- `git mv` the bridge sources into `shared/src/logging/`; legacy `dlt.cc`/`libdlt.cc` stay in the shell.
- `ffi_shim.hpp` → public `ihs/logging.h` (flat entries + `IhsLoggingApi`).
- `logger.hpp` becomes a header-only convenience over the C ABI (no C++ bridge symbol crosses the boundary); `engine.cc`'s engine-log site uses it.
- `-z nodelete` so the worker thread and ring TLS survive a plugin `dlclose`.
- The compat-matrix / load whitebox harnesses compile the relocated sources directly.

## Tracing
- `ihs/trace.h`: `ihs_trace_*` entries + `IhsTraceApi`, zero-cost-when-disabled
  macros (gated on a relaxed `ihs_trace_enabled()`), and an `IHS_TRACE_SCOPE` RAII helper.
- Sink priority: the Flutter engine trace procs (registered by`libflutter_engine.cc` when the proc table resolves) → kernel ftrace `trace_marker` (captures backend bring-up before the engine) → nop.
- `ihs_shared` is now linked into the shell unconditionally (it carries tracing regardless of `ENABLE_DLT`).

## Config
- `ihs/config.h`: an opaque, reference-counted `IhsConfigSnapshot` with a generation counter, per-view and global key scopes, and typed getters. A held snapshot stays valid across a config replacement; a plugin polls the generation to notice a change. `main.cc` publishes a snapshot of the parsed configuration at startup. Keys are flattened names, so extending the set needs no ABI change.

## Verification
- External C11 consumers exercise each surface: the logging sub-table wiring; a fake trace sink receives begin/end/instant with the enable flag tracking set/clear; the config API round-trips every type, verifies type/scope/missing misses and generation increment, and confirms a held snapshot keeps its values after a newer generation is published.
- The config snapshot refcount lifecycle is clean under ASan/UBSan/LeakSanitizer.
- `homescreen` builds with `ENABLE_DLT` on and off; the compat smoke and drop-free load test pass; the shell publishes config without crashing at startup.
- `libihs_shared.so` still exports only the `ihs_*` C ABI.
- clang-tidy-19 (`--warnings-as-errors='*'`) and clang-format-19 clean; both scripts now cover `shared/`.

---

## Hardware validation — riscv64 (StarFive VisionFive 2)

Cross-built for **riscv64** in a `debian:trixie` container (`crossbuild-essential-riscv64`) and validated on a VisionFive 2 (JH7110, Ubuntu 24.04, kernel 6.12.5-starfive) — a second architecture for the C-ABI boundary:

- **ABI smoke on-target** — the out-of-tree `ihs_consumer` (this PR's F2 payload) runs on riscv64 silicon: version handshake, config publish/acquire/read round-trip, trace nop, and DLT self-disable (no `libdlt.so.2` on the board) all pass. The linker version script holds on riscv64 — `libihs_shared.so` exports exactly the **31 `ihs_*` symbols, 0 leaked** internals.
- **Full shell + wonderous (software backend)** — `homescreen` links `libihs_shared.so.1`, boots, loads the riscv64 Flutter engine + a wonderous AOT bundle, **publishes the config snapshot (M5)**, **self-disables DLT (M3)**, opens the DC8200 (`IVI_SW_SINK=drm-dumb:/dev/dri/card0`) and **scans out to HDMI at 3840x2160@30** (`discarded=0`). Wonderous renders (CPU-rastered — slow on this GPU-less board, but correct).
- **All backends cross-compile** — `software`, `wayland-egl`, and `drm-kms-egl` all build clean for riscv64 with the M3/M4/M5 changes. (drm-kms-egl brings up the engine + wonderous correctly, but the board's DC8200 KMS driver drops `PAGE_FLIP_EVENT`s → no smooth scanout; a board/driver quirk unrelated to this change.)

Confirms the C-ABI boundary, hidden-visibility version script, `-z nodelete`, atomics, and the shell's link against `ihs_shared` all work on riscv64.

---

## Update — generic logging surface (review rework)

Per review feedback, the logging ABI is now a **generic** interface rather than
DLT-shaped: `ihs_log_start/stop/flush`, `ihs_log_context_open(tag, options)`,
`ihs_log()` (was `ihs_dlt_*`). DLT is demoted to one **sink** behind the drain
worker, alongside a **ConsoleSink** (stderr, timestamp+level+tag, tty color) and
a **RotatingFileSink** (size-based, no new dependency). Sinks are selected by
`IHS_LOG_SINK` (dlt|console|file), with a warn-once console fallback so a program
never logs silently to nowhere; `IHS_LOG_LEVEL` sets a ring-side floor and
`IHS_LOG_FILE*` configures the file sink. Logging now works with **no
dlt-daemon** (console/file), instead of self-disabling entirely. The abidiff
baseline is re-dumped to freeze the generic surface. DLT context ids move to an
`IhsLogContextOptions` hint.


jwinarske:
Added a 4th commit: **F1/F2 CI gates** (`.github/workflows/ihs-shared.yml`) — an out-of-tree `find_package` consumer smoke (ENABLE_DLT on/off) that also compiles every installed header as C11, and a libabigail `abidiff` ABI gate against a committed baseline (self-bootstrapping on first run). `shared/` now builds standalone so the gates need only a C++ toolchain. Verified locally end-to-end.

PR #320, opened 2026-07-16
Unit Test Coverage Expansion - CI Harness

kerberjg:
Wire the existing `lcov`+GTest harness into CI, re-enable and fix broken tests.

**Goal:** A new GitHub Actions job that builds with `COVERAGE=ON`, runs all test drivers via CTest, captures `lcov`, and uploads the HTML report as an artifact.


kerberjg:
Result:
- all unit tests have been updated & fixed
- **current coverage is at 19%**, report URL [here](https://github.com/toyota-connected/ivi-homescreen/actions/runs/29516028125/artifacts/8382766744)
- CI run time shortened from 15 min+ to 3 min

image

@jwinarske Phase 0 is ready for review - coming up next, phase 1 (new tests)

.github/workflows/oss-ivi-homescreen.yml
```diff
@@ -284,3 +284,79 @@ jobs:
...
 
+  # Unit-test + coverage job.
+  #
+  # Builds the full unit-test suite with -DCOVERAGE=ON (--coverage on every
+  # object), runs all *_test_driver binaries via CTest, captures per-file
+  # line/branch/function data with lcov, filters out third-party and test
+  # sources, generates an HTML report, prints the coverage percentages to the
+  # job log, and uploads the HTML as a downloadable artifact.
+  unit-tests:
+    if: ${{ github.server_url == 'https://github.com' && github.ref != 'refs/heads/agl' && always() }}
```
jwinarske, line 295: Why are we maintaining this line?

.github/workflows/oss-ivi-homescreen.yml
```diff
@@ -284,3 +284,79 @@ jobs:
...
+        run: |
+          echo "=== Coverage Summary ==="
+          lcov --summary coverage.info 2>&1 | grep -E 'lines\.|functions\.|branches\.'
+
+      - name: Generate HTML report
+        working-directory: ${{github.workspace}}/build/unit-tests
+        run: genhtml -o lcovHtml -s --legend --branch-coverage coverage.info
+
+      - name: Upload coverage HTML report
+        uses: actions/upload-artifact@v4
```
jwinarske:
```
Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: actions/upload-artifact@v4...
```
update to v5 to clear this warning

test/unit_test/accessibility_tree-test/CMakeLists.txt
```diff
@@ -4,35 +4,10 @@ set(TESTCASE_NAME "homescreen_accessibility_tree_ut_test_driver")
...
-
-target_link_libraries(
-        ${TESTCASE_NAME}
-        PRIVATE
-        gtest_main
-        ${TYPICAL_TEST_LINK_LIBS}
-        toolchain::toolchain
-)
+target_compile_definitions(${TESTCASE_NAME} PRIVATE TEST_SCRATCH_DIR="${CMAKE_CURRENT_BINARY_DIR}")
+target_link_libraries(${TESTCASE_NAME} PRIVATE gtest_main homescreen_ut_shell)
```
jwinarske, line 9: `toolchain::toolchain` is required to pull in our compiler settings; not optional.  Otherwise we are testing something completely different.

test/unit_test/backing_store_pool-test/CMakeLists.txt
```diff
@@ -3,33 +3,9 @@ set(TESTCASE_NAME "homescreen_backing_store_pool_ut_test_driver")
...
-)
-
-target_link_libraries(
-        ${TESTCASE_NAME}
-        PRIVATE
-        gtest_main
-        ${TYPICAL_TEST_LINK_LIBS}
-        toolchain::toolchain
-)
+target_link_libraries(${TESTCASE_NAME} PRIVATE gtest_main homescreen_ut_shell)
```
jwinarske, line 7: same, missing `toolchain::toolchain` target.  Should also be on all the executables.

test/unit_test/compositor_registry-test/CMakeLists.txt
```diff
@@ -13,33 +13,9 @@ set(TESTCASE_NAME "homescreen_compositor_registry_ut_test_driver")
...
-)
-
-target_link_libraries(
-        ${TESTCASE_NAME}
-        PRIVATE
-        gtest_main
-        ${TYPICAL_TEST_LINK_LIBS}
-        toolchain::toolchain
-)
+target_link_libraries(${TESTCASE_NAME} PRIVATE gtest_main homescreen_ut_shell)
```
jwinarske, line 17: same as above

test/unit_test/compositor_registry-test/CMakeLists.txt
```diff
@@ -13,33 +13,9 @@ set(TESTCASE_NAME "homescreen_compositor_registry_ut_test_driver")
 set(TESTCASE_CC test_case_compositor_registry.cc)
 list(REMOVE_ITEM TYPICAL_TEST_DEFINITIONS "ENABLE_PLUGIN_URL_LAUNCHER")
 
-add_executable(
-        ${TESTCASE_NAME}
-        ${TYPICAL_TEST_SOURCES}
-        ${TESTCASE_CC}
-)
-
+add_executable(${TESTCASE_NAME} ${TESTCASE_CC})
```
jwinarske, line 15: same as above

test/unit_test/configuration-test-case/CMakeLists.txt
```diff
@@ -6,36 +6,13 @@ set(TESTCASE_CC test_case_configuration.cc;
...
 list(REMOVE_ITEM TYPICAL_TEST_DEFINITIONS "ENABLE_PLUGIN_URL_LAUNCHER")
 
-# Basically, the following statements need not be modified
-add_executable(
-        ${TESTCASE_NAME}
-        ${TYPICAL_TEST_SOURCES}
-        ${TESTCASE_CC}
-)
-
+add_executable(${TESTCASE_NAME} ${TESTCASE_CC})
```
jwinarske, line 8: same as above

test/unit_test/configuration-test-case/CMakeLists.txt
```diff
@@ -6,36 +6,13 @@ set(TESTCASE_CC test_case_configuration.cc;
...
-        ${TESTCASE_NAME}
-        PRIVATE
-        gtest_main
-        ${TYPICAL_TEST_LINK_LIBS}
-        toolchain::toolchain
+target_compile_definitions(${TESTCASE_NAME} PRIVATE
+    SOURCE_ROOT_DIR="${CMAKE_CURRENT_SOURCE_DIR}"
+    UNIT_TEST_APP_BUNDLE="${UNIT_TEST_APP_BUNDLE}"
 )
+target_link_libraries(${TESTCASE_NAME} PRIVATE gtest_main homescreen_ut_shell)
```
jwinarske, line 14: same as above

test/unit_test/gl_process_resolver-test/CMakeLists.txt
```diff
@@ -4,35 +4,10 @@ set(TESTCASE_NAME "homescreen_gl_process_resolver_ut_test_driver")
...
-# Basically, the following statements need not be modified
-add_executable(
-        ${TESTCASE_NAME}
-        ${TYPICAL_TEST_SOURCES}
-        ${TESTCASE_CC}
-        ${PROJECT_SOURCE_DIR}/shell/backend/gl_process_resolver.cc
-)
-
+# gl_process_resolver.cc is already in homescreen_ut_shell via TYPICAL_TEST_SOURCES.
+add_executable(${TESTCASE_NAME} ${TESTCASE_CC})
```
jwinarske, line 7: same as above; test with same compiler settings as binary artifact

test/unit_test/libinput_log-test/CMakeLists.txt
```diff
@@ -4,34 +4,26 @@ set(TESTCASE_NAME "homescreen_libinput_log_ut_test_driver")
...
+endif ()
 
-target_link_libraries(
-        ${TESTCASE_NAME}
-        PRIVATE
-        gtest_main
-        ${TYPICAL_TEST_LINK_LIBS}
-        toolchain::toolchain
-)
+add_executable(${TESTCASE_NAME} ${TESTCASE_CC})
```
jwinarske, line 22: same

test/unit_test/mutation_stack-test/CMakeLists.txt
```diff
@@ -3,33 +3,9 @@ set(TESTCASE_NAME "homescreen_mutation_stack_ut_test_driver")
...
-)
-
-target_link_libraries(
-        ${TESTCASE_NAME}
-        PRIVATE
-        gtest_main
-        ${TYPICAL_TEST_LINK_LIBS}
-        toolchain::toolchain
-)
+target_link_libraries(${TESTCASE_NAME} PRIVATE gtest_main homescreen_ut_shell)
```
jwinarske, line 7: same

jwinarske (changes requested):

jwinarske:
@kerberjg Claude's review

1. Your toolchain::toolchain comments are actually already satisfied — transitively. homescreen_ut_shell links toolchain::toolchain PUBLIC (test/unit_test/CMakeLists.txt:96-101), and toolchain is a pure INTERFACE target (cmake/compiler_gnu.cmake:3, compiler_clang.cmake:73). Every test executable linking homescreen_ut_shell inherits its INTERFACE compile/link options — on GNU that's the static-libstdc++/link options; on clang the -stdlib/-isystem flags. The warning flags (-Wall -Wconversion etc.) are directory-scope add_compile_options in cmake/compiler.cmake:52 and hit the test TUs regardless. So the drivers are built with the same compiler settings as the artifact. You could still ask for it explicitly on each executable for greppability, but it's not a correctness gap.
2. Dead REMOVE_ITEM TYPICAL_TEST_DEFINITIONS lines — 16 files. With defines now propagated PUBLIC from the pre-built shell, list(REMOVE_ITEM TYPICAL_TEST_DEFINITIONS "ENABLE_PLUGIN_URL_LAUNCHER") in each per-test CMakeLists mutates a local variable nothing consumes. Two consequences: (a) dead code that implies behavior it doesn't have, and (b) a silent semantic change — utils-test et al. previously compiled shell sources without that define; now the shared objects have it. CI is green so nothing breaks today, but the lines should be deleted and the change acknowledged in the PR description.
3. CI job double-executes the entire suite. The job runs ctest, then lcov -z, then re-runs every *_test_driver in a find loop. Same binaries, same coverage — the reset+rerun adds nothing except doubling test wall time. Keep the ctest step (proper failure reporting), delete "Reset coverage counters" and "Run test drivers (coverage collection)", and capture directly after ctest.
4. always() in the job if: (your open question). It's copy-pasted from the ivi-homescreen/accessibility jobs (lines 17, 127 have the identical expression). With no needs: it does nothing useful and has the known footgun of running even on workflow cancellation. Should be dropped here — and arguably cleaned up in the older jobs in a follow-up.
5. Moved "integration" tests are now dead code. test/integration/{app-test,flutter_view-test,texture-test} exist with CMakeLists that still reference TYPICAL_TEST_SOURCES/TYPICAL_TEST_DEFINITIONS, but nothing adds test/integration as a subdirectory (test/CMakeLists.txt only adds unit_test). They'll bit-rot unbuilt. Either wire them behind a BUILD_INTEGRATION_TESTS option or note explicitly that they're parked.
6. Test-order dependency in utils-test. GetConfigHomePath Lv1Normal002 relies on Lv1Normal001 running first to populate the static cache with the /tmp path — the comment even says "runs first." Under --gtest_shuffle or --gtest_filter, 002 running first caches the fallback path and then 001 fails (its EXPECT_EQ on the /tmp path hits the already-populated cache). 002's own assertions are order-safe; 001 is not. Worth a comment at minimum, or merge them into one test.
Minor: upload-artifact@v4 → v5 (you flagged it); the wake_event_fd.cc explicit append to the shell dedupes against the conditional shell source (identical absolute path, CMake dedupes single-target sources) so no double-compile; --rc lcov_branch_coverage=1 is correct for the lcov 1.x on ubuntu-22.04 but will need renaming to branch_coverage if the runner ever moves to 24.04/lcov 2.x; and the now-recursive submodule checkout in oss-codeql.yml will pull googletest into CodeQL's scan — check whether paths-ignore covers third_party/ or you'll get third-party findings noise.
Verdict: request changes stands, but the substance is items 2, 3, and 5 — item 1 you can resolve as answered. The shell consolidation and submodule migration are solid and I'd want them to land; the 15→3 min CI claim is credible given ~550 redundant TUs eliminated.

jwinarske:
@kerberjg  Add `Job summary ($GITHUB_STEP_SUMMARY)`. Render the totals and worst files as markdown right on the run page instead of buried in the log:
```yaml
yaml- name: Coverage summary
  working-directory: ${{github.workspace}}/build/unit-tests
  run: |
    {
      echo "## Unit test coverage"
      echo '```'
      lcov --summary coverage.info 2>&1 | grep -E 'lines|functions|branches'
      echo '```'
      echo "### Lowest-covered files"
      echo '| File | Lines |'
      echo '|---|---|'
      lcov --list coverage.info 2>/dev/null \
        | awk -F'|' '/%/ {gsub(/ /,"",$2); print $1"|"$2}' \
        | sort -t'|' -k2 -n | head -15 \
        | awk -F'|' '{printf "| %s | %s |\n", $1, $2}'
    } >> "$GITHUB_STEP_SUMMARY"
```

.github/workflows/oss-ivi-homescreen.yml
kerberjg: The logic is "_should run together with `ivi-homescreen` jobs_". Beyond that, the actual rationale for why that's present in `ivi-homescreen` is undocumented, so it seems like maintaining that logic is safer.

If there's a known rationale available for why the original line is necessary and why this logic shouldn't extend to `unit-test`, please let me know

test/unit_test/accessibility_tree-test/CMakeLists.txt
kerberjg: It's included transitively via `homescreen_ut_shell` - see your own comment below (via Claude 1.), it is correct

test/unit_test/backing_store_pool-test/CMakeLists.txt
kerberjg: [same as above](https://github.com/toyota-connected/ivi-homescreen/pull/320#discussion_r3614394941)

kerberjg:
@jwinarske Thank you for the feedback! Answering the points:

1. This is correct
2. 👍
3. I'll have a look and correct as needed
4. See my question above
5. I'll add a script+docs to run them manually since they're not CI-runnable (yet - we can expand on this in the future)
6. I'll have a look at those

>  Add Job summary ($GITHUB_STEP_SUMMARY).

TIL! This is useful, thank you

.github/workflows/oss-ivi-homescreen.yml
jwinarske: It's left over from early `agl` release.  We can drop it.

jwinarske (approved):

PR #328, opened 2026-07-19
wayland-leased-drm: gate direct-evdev input, default off under a host session

jwinarske:
Input hardening for the `wayland-leased-drm` backend family (the input workstream from the leased-drm design).

## The problem

A leased backend has no `wl_surface`, so the compositor delivers it no input — its only source is libinput on `/dev/input/event*`, and those reads are **not** grabbed. On a host with a live Wayland session that means every keystroke and pointer event reaches **both** this process and the session compositor: input is *duplicated, not stolen*. Until now there was no way to stop it — the seat was built and started unconditionally, and only the software tier had any input gate at all (`IVI_SW_INPUT=none`).

## What this adds

A `lease_input` policy, resolved once per leased `make_display`:

- **`auto`** (default) — read evdev on an embedded target, but **disable it and warn** when a host Wayland session is detected (`WAYLAND_DISPLAY` set + socket live).
- **`on`** — always read evdev, even under a session (embedded-under-compositor, where input devices are partitioned between the two).
- **`off`** — never read evdev.

Auto defaults **off** under a detected session so a developer's keyboard is not silently duplicated on a workstation; an embedded target has no `WAYLAND_DISPLAY` and is unaffected. Config surface mirrors the other lease knobs: `--lease-input`, `HOMESCREEN_LEASE_INPUT`, `[view.backend.lease] input`.

## Design

- **Gate mechanism:** `DrmDisplay` and `SoftwareDisplay` gain `SetInputEnabled()`; when false, `StartEvents()` leaves the seat unstarted, so libinput never opens a device (both seats open in `Start()`, not their constructors — verified). Default true, so the direct / path-opened tiers are **unchanged**; only the leased factories set it.
- **Testable policy:** the decision table lives in a dependency-free `leased_input_policy.h` pure function so it is unit-testable without the backend graph; `register_backends` does the config read, session detection, and logging around it. Four cases were added to the existing lease-client test suite (no new test target).

## Correction to a prior assumption

The leased-drm docs claimed `--drm-no-seat` was the dev-host input recipe. It is not — `--drm-no-seat` bypasses *libseat*, not input, and the leased path already runs as `no_seat`. `lease_input` is the actual knob; the README is corrected here.

## Verification

- Builds clean: leased+egl+software, leased+vulkan, and default-Wayland (the shared `DrmDisplay`/`SoftwareDisplay`/`configuration.cc` changes are inert on the direct and non-DRM paths).
- Tests: lease-client suite 30/30 (26 existing + 4 new policy cases), vkms suite 9/9.
- clang-tidy (scoped) applies nothing; clang-format clean; config reference regenerated and in sync.

## Runtime caveat (read before merging)

The full leased-input path **past a granted lease** is not runtime-exercised here — it needs a leasing compositor or root+vkms, neither available in this environment (wlroots offers a connector only when its EDID carries the non-desktop flag; the local session offers none). So the **policy** is proven by the unit tests and the **gate** by construction, not by a live leased frame. This is the same runtime-proof gap the leased GPU tiers already carry; it closes with HIL on real leasing hardware.

### DIFF
--- a/README.md
+++ b/README.md
@@ -303,6 +303,16 @@ Usage:
                               compositor defer the DRM fd until it regains DRM 
                               master, so this is what stops a VT-switched-away 
                               compositor hanging startup.
+      --lease-input arg       wayland-leased-drm: whether to read input from 
+                              raw evdev. A leased client has no wl_surface, so 
+                              libinput on /dev/input/event* is its only input 
+                              source, and it is ungrabbed -- on a host with a 
+                              live Wayland session that duplicates every 
+                              keystroke into this process and the session 
+                              compositor. auto (default) = read evdev on an 
+                              embedded target but disable it when a Wayland 
+                              session is detected; on = always read evdev; off 
+                              = never.
 ```
 
 <!-- END CLI-REFERENCE -->
@@ -363,6 +373,7 @@ annotated all-keys file see
 | `[view.backend.drm]` | `stage_cursor` | `string` | `auto` | `auto\|yes\|no` | drm |
 | `[view.backend.lease]` | `connector` | `string` | `(sole offer)` | `e.g. HDMI-A-1` | leased |
 | `[view.backend.lease]` | `device` | `string` | `(sole device)` | `index or /dev/dri/cardN` | leased |
+| `[view.backend.lease]` | `input` | `string` | `auto` | `auto\|on\|off` | leased |
 | `[view.backend.lease]` | `on_revoke` | `string` | `exit` | `exit\|gate` | leased |
 | `[view.backend.lease]` | `timeout_ms` | `int` | `5000` | `> 0` | leased |
 | `[view.output]` | `drm_connector` | `string` | `(none)` | `e.g. HDMI-A-1` | drm |
--- a/docs/config-examples/reference.toml
+++ b/docs/config-examples/reference.toml
@@ -183,6 +183,10 @@
   # type=string default=(sole device) values=index or /dev/dri/cardN applies=leased
   device =
 
+  # Whether to read input from raw evdev. A leased client has no wl_surface, so ungrabbed libinput is its only input source; under a host Wayland session that duplicates events into both this process and the compositor. auto disables it when a session is detected, on forces it (embedded-under-compositor), off never reads evdev.
+  # type=string default=auto values=auto|on|off applies=leased
+  input =
+
   # On revocation (every VT switch away from the compositor is one): exit non-zero so a supervisor restarts and renegotiates, or gate and keep running with a frozen panel (debugging only).
   # type=string default=exit values=exit|gate applies=leased
   on_revoke =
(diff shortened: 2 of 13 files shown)


jwinarske:
Rebased onto v3.0 and added a test tier, because most of this turned out to be cheap to cover.

**The risk is in the decision, and the decision is already pure.** `DecideLeasedInput(mode, session_present)` takes the session as a `bool` rather than probing the environment, and `leased_input_policy.h` is header-only with no dependency beyond `<string_view>`. So the interesting half needs no compositor, no lease, no DRM and no input devices — only gtest and an include path.

New **T0** tier, 7 cases, runs anywhere:

```
[==========] 7 tests from 1 test suite ran.
[  PASSED  ] 7 tests.
```

Covering: auto disables under a session (the case the feature exists for), auto enables without one, unset behaves as auto so no existing config changes, `on` overrides the session heuristic (or embedded-under-compositor is unreachable), `off` holds where auto would have enabled, a typo falls back to auto **and is reported** rather than silently reading as `off`, and every recognized spelling stays unreported.

**What is still untested** is one hop: that `SetInputEnabled(false)` leaves the seat unstarted so libinput is never read (`drm_display.cc:469`). That is a two-line effect, and it is the only part needing a machine with a real compositor and real devices.

Also verified on the rebase: T1 (mock drm-lease-v1 compositor) still passes 30 tests, the generator stays reproducible — re-running `gen_config_reference.py` reports "nothing to update" — and the branch builds against v3.0 after 296 commits of drift.

jwinarske:
Both addressed, and the second one caught a mistake of mine.

**The extra test target was redundant — removed.** The reviewer is right that the PR adds a target, and right to question it: this commit *already* tests the policy inside `lease_client-test` (`test_case_lease_client.cc:861-901`). I added a separate tier yesterday without checking that file's contents — I read the diff's file list, saw `leased_input_policy.h` as new, and assumed the decision was uncovered. It was not. Consolidated as suggested: the separate target and directory are gone, and the branch is back to a single test target.

Checking what the existing suite covers, it was already thorough — `""` and typos in both session states (`:882-894`), and the negative case for `auto`/`""`/`off`/`on` (`:896-901`). Only two assertions were genuinely additive, so those were folded into the existing case rather than a new one:

- every accepted spelling (`yes`/`true`/`1`/`none`/`false`/`0`) is recognized, so a correct config never warns;
- matching is **case-sensitive** — `"Off"` is reported and treated as auto rather than silently disabling input. That is the safer way to be wrong, but it is a real way to mis-write the key, so the behavior is now pinned.

`lease_client-test`: 30 tests, all passing.

**Comment formatting fixed.** The stray `"a"` line in the `lease_input` doc block (`configuration.h:141`) was a bad wrap; rewrapped.

The PR description's "no new test target" is accurate again as a result, so no description change is needed.

PR #330, opened 2026-07-21
drm: scan out platform-view dma-bufs on KMS overlay planes

jwinarske:
## What

On drm-kms-egl the DRM compositor previously forced any frame containing a
platform view through the GL composite path, which assembled the frame but never
placed the view — an `ihs_pv`-submitted dma-buf never reached scanout. This
routes it onto a KMS overlay plane instead.

- `ICompositorSurface` gains a `Dmabuf` accessor (`GetDmabuf`); the ihs_pv host's
  `IhsPluginView` implements it as a pure read of the frame it already retains
  from the last submit. The compositor plane-routes XOR GL-imports a surface per
  frame so the two paths never both consume the frame.
- `PresentLayersViaScene` wraps each platform view's dma-buf in an
  `ExternalDmaBufSource` scene layer at the Flutter layer rect, and writes the
  Flutter-array position as the plane zpos on every non-primary layer so the
  overlay stack shares one z scheme. A view with no dma-buf, or a layer the
  allocator can't place, falls back to GL composition.
- The EGL context now carries `gbm_device`, so a GL/dma-buf producer on a DRM
  backend can allocate a scanout-capable buffer.

## Dependencies

Bumps drm-cxx to **v2.0.1** for the allocator warm-start stability fix — it keeps
each view on its plane across Flutter's per-frame backing-store re-splits;
without it, mousing over reshuffles the overlay planes (flicker).

Pairs with the in-tree ihs_pv producer in ivi-homescreen-plugins that exercises
this path (linked below).

## Validation

HW-validated on Raspberry Pi 4 (vc4), drm-kms-egl: solid-color platform-view
boxes direct-scanout onto KMS overlay planes (`LayerScene commit total=14
assigned=14 composited=0`), zero GL fallback across 146 interactive frames,
stable and flicker-free under mouse interaction.

### DIFF
--- a/shell/backend/backend.h
+++ b/shell/backend/backend.h
@@ -48,6 +48,9 @@ struct BackendEglContext {
   void* share_context;  // EGLContext suitable as share_context for a
                         // plugin-owned context that uploads GL objects
                         // visible on the raster thread.
+  void* gbm_device;     // struct gbm_device* on a gbm/DRM backend (NULL on a
+                        // Wayland EGL backend). Lets a platform view allocate a
+                        // scanout-capable buffer it can export as a dma-buf.
 };
 
 // Carries the Vulkan handle set a plugin needs to render into a VkImage using
--- a/shell/backend/drm_kms_egl/drm_backend.cc
+++ b/shell/backend/drm_kms_egl/drm_backend.cc
@@ -1988,5 +1988,6 @@ bool DrmBackend::GetEglContext(BackendEglContext* out) const {
   out->display = egl_display_;
   out->config = egl_config_;
   out->share_context = egl_context_;
+  out->gbm_device = gbm_device_;
   return true;
 }
--- a/shell/backend/drm_kms_egl/drm_compositor.cc
+++ b/shell/backend/drm_kms_egl/drm_compositor.cc
@@ -2397,11 +2397,18 @@ bool DrmCompositor::PresentLayersViaScene(const FlutterLayer** layers,
       // reached it yet).
       surface->OnResize(static_cast<int32_t>(fl->size.width),
                         static_cast<int32_t>(fl->size.height));
-      ICompositorSurface::Dmabuf db;
-      if (!surface->GetDmabuf(&db)) {
-        // GL-only surface, or no dma-buf frame ready this present: route
-        // the whole frame through GL composition (which samples the
-        // surface's GL texture / Vulkan image).
+      // Only pull (and dup) a dma-buf for a NEW platform view — one that will
+      // build an ExternalDmaBufSource below. An existing scene layer keeps its
+      // source (the reuse path in the add loop), so calling GetDmabuf for it
+      // every frame would dup+close an fd for nothing. find_by_identity_tag
+      // reflects last frame's layers; this PV is present, so it survives the
+      // prune and the add loop finds it too.
+      ICompositorSurface::Dmabuf db{};
+      if (scene_->find_by_identity_tag(surface.get()) == nullptr &&
+          !surface->GetDmabuf(&db)) {
+        // New platform view with no dma-buf (GL-only surface, or none ready
+        // this present): route the whole frame through GL composition, which
+        // samples the surface's GL texture / Vulkan image.
         return PresentViaGlFallback(layers, layer_count);
       }
       frame_layers.push_back({fl, nullptr, nullptr, surface.get(), db});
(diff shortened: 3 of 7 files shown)


jwinarske:
Producer side: https://github.com/toyota-connected/ivi-homescreen-plugins/pull/260

jwinarske:
Addressed in `a80777ea` (smoke-tested on Pi 4 / vc4 — boxes still direct-scanout on KMS planes, 189 scene layers cycled with no EBADF):
- **[#4] borrowed-fd race** — `GetDmabuf` now dups the fd under its lock and returns an owned copy; the compositor closes it on every exit path via an RAII guard (`ExternalDmaBufSource::create` dups again for the KMS FB). Real race under a continuous producer; fixed.
- **[#1] plane_count overrun** — `GetDmabuf` clamps `plane_count` to the 4 planes the descriptor holds, so the `ExternalPlaneInfo` span can't exceed the array.
- **[#3] acquire_fence_fd contract** — softened the header doc; the DRM scene path is implicit-sync today and doesn't consume it, so `GetDmabuf` leaves it -1.
- **[#2] z_index vs layers[] index** — no functional change: `frame_layers` is built in `layers[]` order and zpos only needs relative ordering (which is preserved), and layer 0 is always the base, so `is_ui` is correct. Left as-is.

jwinarske:
Addressed in `7819be4a` (compiles + clang-tidy + clang-format clean on the drm-kms-egl build):

- **[1–4] overlay zpos ignores non-zero primary zpos** — real bug. Non-UI layers (BS + PV) now use `primary_zpos_ + z_index` instead of the raw index, so on drivers where the primary plane's zpos isn't 0 (amdgpu = 2) overlays stay above it. `primary_zpos_` is already resolved from the primary plane's `zpos_min`, and the composition-buffer path already uses `primary_zpos_ + 1`.
- **[5] plane_count==0 → 1 default** — removed. `GetDmabuf` guarantees a 1..4-plane single-handle frame (returns false otherwise), so the compositor uses `db.plane_count` directly; no zero-offset/stride can reach `ExternalDmaBufSource::create`.
- **[6] existing PV layer not updated to the current dma-buf** — correct observation; documented as a scoped limitation. The current producer's buffer content is static once submitted (a resize just re-stretches a solid color, no visible change), so keeping the `ExternalDmaBufSource` is fine here. A content-changing producer (decoder/camera) needs `LayerScene::replace_source()` driven by a buffer ring — that's the dynamic-producer follow-up, and the code comment now says so explicitly.

jwinarske:
Addressed in `52082943`:
- **single-handle dma-buf assumption** — `GetDmabuf` now validates that every plane references `plane_fd[0]` (the single-handle layout the `Dmabuf` descriptor models); a genuine per-plane-fd frame returns false and takes the GL import path, so no invalid plane descriptor is built.
- **modifier doc** — corrected: `DRM_FORMAT_MOD_LINEAR` (0) is the linear buffer; `DRM_FORMAT_MOD_INVALID` is the unknown/unspecified sentinel (was wrongly described as 'linear').

jwinarske:
Addressed in `a542fda8`:
- **GetDmabuf across a GL-import fallback** — valid; documented the contract. A GL fallback consumes the stashed frame (GetGlTextureName import), so GetDmabuf returns false until the producer resubmits: a continuous producer resumes scanout on its next submit, a static producer stays GL-composited (still correct) until it resubmits. Low impact here (vc4 has 18 overlay planes; boxes stay on planes, so the fallback is rare, and a solid-color box GL-composites identically). Retaining the dma-buf across a GL import so a static producer auto-returns to scanout is folded into the dynamic-producer follow-up.
- The suppressed low-confidence note on drm_compositor.cc:2516 (existing PV layer keeps its ExternalDmaBufSource) is the same point already addressed in `7819be4a` — documented as the replace_source/ring dynamic-producer follow-up.

jwinarske:
Addressed in `c5f42af7`:
- **per-frame fd churn for existing PVs** — good catch. GetDmabuf (dup) is now called only for a *new* platform view: the collection loop checks `scene_->find_by_identity_tag(surface)` first and skips the pull for an existing scene layer (which keeps its source). Steady-state frames no longer dup+close an fd per PV.
- **z_index comment (suppressed 2446)** — reworded to say it's the engaged-layer position in `frame_layers` (Flutter order, compacted), which preserves the relative z-order that matters for stacking.
- **existing PV keeps source (suppressed 2508)** — same dynamic-producer point already documented in `7819be4a`/`a542fda8` (replace_source/ring follow-up).

Issue #208, opened 2026-06-01
Wayland-Vulkan: VkQueue used concurrently by Flutter engine and present path (UNASSIGNED-Threading-MultipleThreads-Write)


jwinarske:
## Summary

On the **Wayland-Vulkan** backend, the Flutter engine and the homescreen present path both submit to the **same `VkQueue`** from **different threads**, with no host synchronization between them. Under `-d` (synchronization + thread-safety validation) this is reported as:

```
UNASSIGNED-Threading-MultipleThreads-Write
vkQueueSubmit()/vkQueuePresentKHR(): THREADING ERROR : object of type VkQueue is
simultaneously used in current thread <A> and thread <B>
```

Vulkan requires host access to a `VkQueue` to be externally synchronized; concurrent use from two threads is undefined behavior, not just a validation nag.

## Root cause

`WaylandVulkanBackend::GetRenderConfig()` hands the engine the backend's own queue:

```cpp
// shell/backend/wayland_vulkan/wayland_vulkan.cc:95-96
.queue_family_index = queue_family_index_,
.queue              = queue_,
```

So:
- The **Flutter raster thread** submits render / texture-upload work on `queue_` (the engine's own `vkQueueSubmit`s — the backend issues no draw/sampler/descriptor calls itself).
- The **platform thread** submits the compositor present (`PresentLayersImpl` / `PresentCallback`) and the one-shot transfer helpers on the same `queue_`.

The backend's `queue_mutex_` serializes the backend's *own* queue operations, but the engine submits directly via the handle it was given and has no knowledge of that mutex, so the two can overlap.

## Reproduction

`BUILD_BACKEND_WAYLAND_VULKAN=ON`, run any bundle with `-d`, then navigate to a texture-heavy screen (e.g. wonderous `tajMahal` page) so the engine streams textures while the compositor presents:

```
homescreen -d -b /path/to/.desktop-homescreen
# observe UNASSIGNED-Threading-MultipleThreads-Write on vkQueueSubmit / vkQueuePresentKHR
```

Occurs with `BUILD_COMPOSITOR` both on and off (both present paths share `queue_`).

## Notes / scope

- This is **separate from #206's black screen** (now fixed in #207). #207 made the compositor present path spec-valid and clears the per-frame render errors on a static screen; this threading issue is about *who* submits to the queue, not *what* is submitted, and surfaces under concurrent engine + present load.
[...]

PR #333, opened 2026-07-22
Vulkan backends: enable device features + expose the device for platform-view engines

jwinarske:
## What

Enable the Vulkan device features that a platform-view engine adopting the backend's shared device needs, on **both** Vulkan backends, and expose the device to the ihs_pv platform-view host on the DRM/KMS Vulkan backend.

Motivated by integrating the Mapbox Maps Embedded SDK as an `ihs_pv` platform view (renders headlessly into an exported dma-buf the compositor imports zero-copy), but the changes are engine-agnostic — any platform-view engine that adopts the shared `VkDevice` benefits.

## Commits

1. **wayland_vulkan: enable device features platform-view engines need** — enable `samplerAnisotropy`, `dualSrcBlend`, and `VK_EXT_pipeline_creation_cache_control` on the logical device when the physical device advertises them. Without these, an adopting engine's anisotropic samplers, dual-source blending, and externally-synchronized pipeline cache trip validation errors.
2. **drm_kms_vulkan: enable device features platform-view engines need** — the same three features on the DRM/KMS Vulkan backend, matching its feature-chain style.
3. **drm_kms_vulkan: expose the Vulkan device to the platform-view host** — implement `VulkanDrmBackend::GetVulkanContext` (mirroring `WaylandVulkanBackend`). Without it the ihs_pv host reported "vulkan off" and every dma-buf producer went inert; with it a plugin adopts the shared device, renders into an exported dma-buf, and the compositor imports and composites it.

## Safety

Every feature is enabled **only when the physical device advertises it** — `samplerAnisotropy`/`dualSrcBlend` are read from the already-queried `features2`; the cache-control extension is appended only when both its feature bit and the extension are present. A device lacking any of them is unaffected (no refusal, no behavior change for other backends or plugins).

`GetVulkanContext` uses the raw instance-proc loader for now (no shared-queue interposer on the DRM Vulkan backend yet — issue #208); a follow-up can add the locking trampolines the Wayland Vulkan backend uses.

## Validation

- **wayland_vulkan** — HW-validated on amdgpu (RADV): a quad of Mapbox map platform views renders tiles, steady-state 0 VUIDs.
- **drm_kms_vulkan** — validated headlessly on vkms: the quad's four producers come up (host reports "vulkan ready"), render, and present to the KMS plane. Real-hardware KMS validation to follow.

### DIFF
--- a/shell/backend/drm_kms_vulkan/vulkan_drm_backend.cc
+++ b/shell/backend/drm_kms_vulkan/vulkan_drm_backend.cc
@@ -1836,11 +1836,17 @@ bool VulkanDrmBackend::GetVulkanContext(BackendVulkanContext* out) const {
   out->device = device_;
   out->queue = graphics_queue_;
   out->queue_family_index = graphics_queue_family_;
-  // Raw loader: this backend does not yet interpose vkQueue* through a shared-
-  // queue lock the way WaylandVulkanBackend does (QueueInterposer), so a
-  // plugin's submits and the compositor's share the one graphics queue without
-  // serialization (issue #208) — benign in practice; an interposed loader is a
-  // follow-up.
+  // KNOWN CONTRACT GAP (issue #208): this hands back the raw loader, but
+  // ihs/platform_view.h requires get_instance_proc_addr to be an INTERPOSED
+  // loader that serializes vkQueueSubmit/vkQueuePresentKHR on the shared
+  // graphics queue (as WaylandVulkanBackend does via QueueInterposer). This
+  // backend has no queue mutex yet, so a plugin resolving through this loader
+  // would submit unsynchronized against the compositor/engine — undefined
+  // behavior per the Vulkan external-synchronization rules. Latent today (no
+  // Vulkan-on-DRM plugin is wired to this ABI), but it MUST be interposed
+  // before one is. Fixing it means adding a queue mutex, wrapping this
+  // backend's own submit/present sites in it, and returning the interposed
+  // loader here and to the engine — tracked under #208.
   out->get_instance_proc_addr =
       reinterpret_cast<void*>(d().vkGetInstanceProcAddr);
   out->device_extensions = enabled_device_extensions_.data();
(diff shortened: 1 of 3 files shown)


jwinarske:
Thanks for the review — addressed one, deferring the other with rationale:

**1. `pipelineCreationCacheControl` feature bit (wayland_vulkan.cc) — fixed** (`26c74ebf`). Now queries `VkPhysicalDevicePipelineCreationCacheControlFeatures` via `vkGetPhysicalDeviceFeatures2` and only chains/enables it when the device reports support, so an advertised-but-unsupported extension can't fail `vkCreateDevice`. This mirrors the query `drm_kms_vulkan` already performs. Compiles clean; clang-tidy-19 + clang-format-19 clean.

**2. Interposed loader for `drm_kms_vulkan::GetVulkanContext` — deferred (tracked, #208).** Good observation, but the fix is larger than it looks and out of scope for this "enable device features / expose the device" PR:
- `BackendVulkanContext` carries no `queue_lock`/`queue_unlock`; the interposed loader is the *only* shared-queue safety mechanism. `WaylandVulkanBackend` has a `queue_mutex_` and serializes its own submits on it before handing out the interposed loader.
- `drm_kms_vulkan` has **no queue mutex at all** yet. Returning an interposed loader without also wrapping this backend's own submit/present sites in the same mutex would serialize only the engine's calls and give a false sense of safety.

So this is really "implement the #208 QueueInterposer mitigation for `drm_kms_vulkan`" — a concurrency change that also can't be exercised without Vulkan-capable KMS hardware (Pi 5). It's already called out in the code (`vulkan_drm_backend.cc:1839-1843`) and tracked under #208. I'll do it as a dedicated follow-up with HW validation rather than fold an untestable concurrency change into this PR.

jwinarske:
Follow-up / correction to my earlier deferral note on the interposed loader.

The low-confidence suppressed comment on the re-review is actually right, and I want to correct something I got wrong above. I checked `ihs/platform_view.h` and the ABI is explicit:

- `:231` — "The plugin MUST resolve every Vulkan function through `get_instance_proc_addr`. It is the same **interposed loader** the engine is [handed]"
- `:250-252` — "Per-submit serialization is handled for you by the interposed loader; use `queue_lock`/`queue_unlock` **only** to make a multi-call sequence atomic"
- `:264` — `get_instance_proc_addr; /* interposed PFN_vkGetInstanceProcAddr */`

So the interposed loader is the contract, and `queue_lock`/`queue_unlock` are not the primary sync mechanism (as I implied). `drm_kms_vulkan` returning the raw loader is a genuine **ABI contract gap**, not "benign in practice" — latent only because no Vulkan-on-DRM plugin is wired to this ABI yet.

The implementation still defers to #208 (it needs a queue mutex + wrapping this backend's own submit/present sites + returning the interposed loader, and can't be HW-validated without a Pi 5). But I've corrected the misleading in-code comment (`0e1eb258`) to state it as a known contract gap that MUST be interposed before a Vulkan-on-DRM plugin is wired, rather than "benign." Thanks for pushing on this.

PR #334, opened 2026-07-22
drm-kms-egl: wire explicit sync for platform-view scanout

jwinarske:
## Summary

Wires explicit synchronization for platform-view direct scanout on `drm_kms_egl`. A producer's acquire fence now rides the overlay plane's `IN_FENCE_FD`, so the display waits on the producer's GPU work instead of the producer stalling on a CPU fence before submit — removing the per-frame stall that showed up as juddery pans with a live map producer.

## Changes

- **Host** (`platform_view_host.cc`): stash the producer's acquire fence with the pending frame instead of closing it, and forward it to the DRM scene path via `GetDmabuf`. The GL-texture fallback (no explicit plane sync) poll-waits on the fence before sampling, then releases it.
- **Compositor** (`drm_compositor.cc`): hand the fence to `ExternalDmaBufSource::set_acquire_fence` on both the new-layer and reused-layer paths (`import_fd` dups, so `PvFdCloser` still closes the original).
- **drm-cxx** bump to include the matching `acquire()` fix ([drm-cxx#230](https://github.com/jwinarske/drm-cxx/pull/230), merged): `acquire()` now hands back a dup of the fence rather than moving it out, so it survives the scene's test + real commit passes and actually reaches the plane's `IN_FENCE_FD`.

## Validation

HW-validated on a Raspberry Pi 4 (vc4) with a MapLibre GL dma-buf producer through `drm_kms_egl`, rebased onto current `v2.0` (incl. #331):

- `DRM_CXX_FENCE_DEBUG=1` → `in_fences_armed=1 in_fence_cpu_waits=0` (the acquire fence reaches the plane's `IN_FENCE_FD`, not a CPU-wait fallback)
- 0 errors, steady scanout, on-panel pans smooth and consistent
- `clang-tidy-19` clean on both changed TUs (host build with all backends); `clang-format-19` clean

## Follow-ups

- Release-fence backpressure (producer↔compositor) — currently a fixed-depth ring
- Producer vsync pacing

### DIFF
--- a/shell/backend/drm_kms_egl/drm_compositor.cc
+++ b/shell/backend/drm_kms_egl/drm_compositor.cc
@@ -40,6 +40,7 @@
 
 #include <drm-cxx/core/format.hpp>
 #include <drm-cxx/scene/external_dma_buf_source.hpp>
+#include <drm-cxx/sync/fence.hpp>
 
 #include "backend/drm_kms_egl/driver_probe.h"
 #include "backend/drm_kms_egl/drm_backend.h"
@@ -2361,6 +2362,16 @@ bool DrmCompositor::PresentLayersViaScene(const FlutterLayer** layers,
           ::close(fl.pv_db.fd);
           fl.pv_db.fd = -1;
         }
+        // Close the producer's acquire fence. drm::sync::SyncFence::import_fd
+        // (used by set_acquire_fence below) DUPS the fd rather than taking
+        // ownership, so this original is always ours to close — the same
+        // pattern vulkan_drm_backend.cc follows (it also ::close()s its fd
+        // after import_fd). Also covers the paths where the source failed to
+        // build.
+        if (fl.pv_tag != nullptr && fl.pv_db.acquire_fence_fd >= 0) {
+          ::close(fl.pv_db.acquire_fence_fd);
+          fl.pv_db.acquire_fence_fd = -1;
+        }
       }
     }
   } pv_fd_closer{frame_layers};
@@ -2539,6 +2550,22 @@ bool DrmCompositor::PresentLayersViaScene(const FlutterLayer** layers,
               drm::span<const drm::scene::ExternalPlaneInfo>(planes.data(),
                                                              np));
           if (src) {
+            // Explicit sync: the producer's acquire fence rides the plane's
+            // IN_FENCE_FD (scene CPU-waits as a fallback). import_fd dups, so
+            // PvFdCloser still closes our original fd.
+            if (db.acquire_fence_fd >= 0) {
+              if (auto f = drm::sync::SyncFence::import_fd(db.acquire_fence_fd);
+                  f) {
+                src.value()->set_acquire_fence(std::move(f.value()));
+              } else {
+                // Fall back to implicit sync for this frame (the producer
+                // synced, or scanout may tear); log so it's diagnosable.
+                ihs::log::warn(
+                    "[DrmCompositor] pv acquire-fence import failed ({}); "
+                    "implicit sync this frame",
+                    f.error().message());
+              }
+            }
             if (auto r = scene_->replace_source(layer->handle(),
                                                 std::move(src.value()));
                 !r) {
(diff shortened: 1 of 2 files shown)


jwinarske:
Addressed the review comments (force-pushed):

1. **GetDmabuf ownership (host:330)** — now returns a `dup` of the acquire fence and retains the surface's copy, per the `ICompositorSurface::Dmabuf` contract. This also fixes the fallback correctness: if a present falls back to `PresentViaGlFallback` after `GetDmabuf`, `GetGlTextureName` can still wait on the fence. The retained fd is closed on the next superseding submit, on dispose, or by the fallback wait.
2. **GL-fallback poll result (host:484)** — capture `poll()`'s result and, on timeout/error, log a warning before best-effort sampling (a wedged fence can't hang the raster thread, but it's now diagnosable).
3. **Stale comment (host:763)** — updated to state the GL-texture fallback CPU-waits on the fence (no longer "ignored").
4. & 5. **`import_fd` failure (compositor:2558/2608)** — both sites now log a warning and fall back to implicit sync for that frame instead of silently dropping the fence.

Re-validated on a Pi 4 (vc4): `DRM_CXX_FENCE_DEBUG=1` → `in_fences_armed=1 in_fence_cpu_waits=0`, 0 errors, smooth scanout. clang-tidy-19 + clang-format-19 clean on both files.

jwinarske:
Addressed the latest review comments (force-pushed):

1. **`dup()` failure (host:335)** — `GetDmabuf` now checks the `dup()` result; on failure it leaves `acquire_fence_fd = -1` (implicit sync for that frame) and logs a warning with `errno`.
2. **Lock held during poll-wait (host:498)** — `GetGlTextureName` now uses a `unique_lock` and drops it across the (bounded) fence wait so it can't block `HostSubmit` or dispose. It takes ownership of the fd before unlocking (so a superseding submit can't close it), loops to also wait on any frame that arrived while unlocked (never sampling ahead of the producer), and bails safely if the view was disposed during the wait.

Re-validated on a Pi 4 (vc4): `in_fences_armed=1 in_fence_cpu_waits=0`, 0 errors, smooth. clang-tidy-19 + clang-format-19 clean.

jwinarske:
On the three `import_fd` ownership comments (drm_compositor.cc:2367/2565/2621): the premise that `SyncFence::import_fd` takes ownership of the fd is not correct — it **dups**, so the current code (closing the original in `PvFdCloser`) is right, and clearing `fl.pv_db.acquire_fence_fd` after import would **leak** the compositor's dup.

Evidence, from the same file the review cites:

- `drm-cxx/src/sync/fence.cpp` — `SyncFence::import_fd`:
  ```cpp
  int const duped = ::dup(fence_fd);
  ...
  return SyncFence(duped);
  ```
- `shell/backend/drm_kms_vulkan/vulkan_drm_backend.cc:1378-1381` (the cited call site) actually **closes the original after import_fd**, which confirms the dup semantics:
  ```cpp
  if (auto fence = drm::sync::SyncFence::import_fd(scanout_fence_fd)) { ... }
  ::close(scanout_fence_fd);
  ```

There are three independently-owned fds, each closed exactly once: the surface's original (host, on supersede/dispose), the compositor's dup (`fl.pv_db.acquire_fence_fd` from `GetDmabuf`, closed by `PvFdCloser`), and the `SyncFence`'s dup (closed on buffer release). No double-close or premature-close.

I did tighten the `PvFdCloser` comment (`ae0e51b3`) to state explicitly that `import_fd` dups (referencing the `vulkan_drm_backend.cc` precedent) so this ownership is unambiguous to future readers.

jwinarske:
Re the latest re-review's low-confidence notes (`562e1c2f`):

Two were valid and are fixed:
- **`PendingEglFrame` comment (host:183)** — reworded; it now says GetDmabuf hands the scene path a *dup* and this copy stays owned here (closed on superseding submit / dispose / the GL-fallback wait), instead of "until GetDmabuf takes it."
- **poll() revents (host:504)** — now treats only `POLLIN` as a signal; `poll()` can wake with `POLLERR`/`POLLHUP`/`POLLNVAL` (>0 without POLLIN), which is a failure, so the warn condition is `pr <= 0 || !(revents & POLLIN)` and the log includes `revents`.

The other three notes (drm_compositor.cc:2374/2567/2623) restate that `SyncFence::import_fd` takes ownership — it doesn't, it dups (`fence.cpp`: `int const duped = ::dup(fence_fd)`), and the cited `vulkan_drm_backend.cc:1381` closes its original *after* import_fd, which confirms dup semantics. Clearing `fl.pv_db.acquire_fence_fd` on success (as suggested) would leak the compositor's dup, so no change there — see my earlier reply with the code refs.

Re-validated on a Pi 4: `in_fences_armed=1 in_fence_cpu_waits=0`, 0 errors. clang-tidy-19 + clang-format-19 clean.

PR #335, opened 2026-07-23
platform-views: sample planar YUV through an external image

jwinarske:
## The problem

`EglDmabufImporter::Import` bound every dma-buf to `GL_TEXTURE_2D`, and `GlCompositor` sampled it with a `sampler2D`. For a planar YUV image that is the wrong target: the driver exposes the first plane alone and applies no colour conversion, so the frame arrives as `(Y, 0, 0, 1)`.

A hardware-decoded NV12 video frame reaching a platform view therefore rendered as a red luma ramp — geometry, stride and motion all correct, colour absent.

`egl_dmabuf_import.h` already recorded the gap: *"planar YUV → GL_TEXTURE_EXTERNAL_OES is a follow-up."* This is that follow-up.

## The change

- `IsPlanarYuv()` covers NV12/NV21/NV16/NV61, YUV420/YVU420/YUV422/YUV444, P010, YUYV, UYVY. Those import into `GL_TEXTURE_EXTERNAL_OES`; everything else keeps `GL_TEXTURE_2D`.
- `ImportedTexture::external` records which, surfaced as `ICompositorSurface::TextureIsExternalOes()` alongside the existing `TextureIsTopFirst()`.
- `GlCompositor` gains a `samplerExternalOES` program and selects it per draw.
- Wired for `wayland-egl` and for all three platform-view composite paths in `drm_compositor.cc`.

Three details worth review attention:

**Optional by construction.** The external program is built only when the context advertises `GL_OES_EGL_image_external`; otherwise `program_external_` stays 0 and a planar frame is left undrawn rather than mis-sampled. Packed formats are unaffected either way.

**No FBO path.** An external texture cannot be attached to an FBO, so both backends skip the blit fast path when `external` is set.

**Not batched.** The external draw sets up and tears down its own state. It has its own attribute and uniform locations and cannot ride the persistent quad state emitted for the `sampler2D` program, so it is deliberately outside that fast path — one platform-view layer per frame, so the saving would not have been meaningful.

The `external` flag is a **trailing default parameter** on `CompositeToFbo`/`CompositeToDefault`/`CompositeViaQuad`/`CompositeLayerIntoFbo`. That keeps every existing call site untouched, at the cost that a new caller can silently omit it. Happy to make it explicit instead if you would rather.

## Testing

A hardware-decoded WebRTC H.264 stream delivered to a platform view as NV12 dma-bufs, via an out-of-tree plugin implementing `IhsPvFactory`:

| | before | after |
|---|---|---|
| AMD desktop, `wayland-egl` | red luma ramp | correct greyscale |
| Raspberry Pi 4, `drm-kms-egl` | — | correct greyscale |

The Pi decodes through `bcm2835-codec` (`/dev/video10`) straight to dma-bufs, so the frame reaches the compositor without a copy.

## Out of scope

`DmabufVulkanImporter` needs a `VkSamplerYcbcrConversion` for the same reason; a planar frame on a Vulkan backend still samples as luma. Not addressed here.

## Two things noticed while working on this, not fixed here

- `HostSubmit` copies the caller's frame with `pending_egl.frame = *frame` without consulting `frame->struct_size`. A plugin built against an older header has a smaller object, so that reads past the end of it — and what lands in the trailing `buffer_id` is then used as the import-cache key. The field exists for exactly this case.
- `*out_release_fence_fd = -1` is still marked a follow-up. A producer that would retire buffers on the release fence instead falls back to a depth window, holding more of its pool than it needs to.

Happy to open issues for either.

### DIFF
--- a/shell/backend/drm_kms_egl/drm_compositor.cc
+++ b/shell/backend/drm_kms_egl/drm_compositor.cc
@@ -999,9 +999,11 @@ void DrmCompositor::CompositeLayerIntoFbo(GLuint target_fbo,
                                           GLsizei dst_w,
                                           GLsizei dst_h,
                                           bool blend,
-                                          bool flip_y) const {
+                                          bool flip_y,
+                                          bool external) const {
   gl_compositor_->CompositeToFbo(target_fbo, src_fbo, src_tex, src_w, src_h,
-                                 dst_x, dst_y, dst_w, dst_h, blend, flip_y);
+                                 dst_x, dst_y, dst_w, dst_h, blend, flip_y,
+                                 external);
 }
 
 // ─── GL fallback (no plane allocator) ────────────────────────────────────
@@ -1073,8 +1075,9 @@ bool DrmCompositor::PresentViaGlFallback(const FlutterLayer** layers,
           // GlFallback writes into FBO 0 (gbm_surface, standard GL NDC).
           // Flip only when the surface reports top-first storage.
           const bool flip_y = surface_sp->TextureIsTopFirst();
+          const bool external = surface_sp->TextureIsExternalOes();
           gl_compositor_->CompositeToDefault(0, tex, sw, sh, dx, dy, dw, dh,
-                                             blend, flip_y);
+                                             blend, flip_y, external);
           composited_any = true;
         }
       }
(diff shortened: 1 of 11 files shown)


jwinarske:
Thanks — all five were real. Fixed in the follow-up commit.

**Extension matched exactly, not by substring.** `GL_OES_EGL_image_external_essl3` would have satisfied a substring test for the base extension, and the shader would then have failed to link with noise rather than the path quietly staying off. Both call sites now use the same space-delimited test `GlCaps::Probe()` already had.

**Importer verifies the extension too — and before closing the fds.** That ordering detail was the important part of the note: doing it after would have returned failure with the fds already closed, and the contract says a failed import leaves them to the caller, so they would have been closed twice. The check now sits immediately before the close loop, destroys the `EGLImageKHR`, and returns false.

**Baseline state.** Depth, stencil, scissor and cull are now disabled in `CompositeViaQuadExternal` too. I had described that path as self-contained and it was not.

**Blend no longer leaks.** It is disabled at the end of the external draw rather than left as the call asked for. Since this path clears `persistent_state_emitted_`, `EndFrame()` tears nothing down, so an enabled blend would have carried into the HUD and the next frame.

**`program_external_` deleted in the destructor.**

Re-verified after the changes on `wayland-egl` with a hardware-decoded NV12 stream: zero import errors, no extension warning, picture unchanged.

jwinarske:
Thanks — all four addressed in 9876e5a0.

**Data race fixed.** `TextureIsExternalOes()` now takes the mutex, matching the sibling `GetGlTextureWidth/Height` accessors. `current_egl` is mutated under lock in `GetGlTextureName`, so the unlocked read was indeed racing the compositor thread.

**Docs corrected, all three sites.** `IsPlanarYuv()` routes packed YUV (`YUYV`/`UYVY`) through `GL_TEXTURE_EXTERNAL_OES` too, so "packed formats stay on `GL_TEXTURE_2D`" was wrong. Reworded `TextureIsExternalOes()` (compositor_surface_interface.h), `ImportedTexture::external` and `Import()` (egl_dmabuf_import.h) to say YUV — planar *or* packed — goes external, and only packed **RGB** stays on `GL_TEXTURE_2D`.

**Includes added.** `gl_compositor.cc` now includes `<cstring>`/`<string>`/`<string_view>` that `ExtensionSupported` uses directly, instead of relying on transitive includes.

Rebuilt on the compositor config; the three touched TUs compile and the shell links.

jwinarske:
Addressed the latest round in the follow-up commit.

**Renamed `IsPlanarYuv()` → `NeedsExternalOes()`** — good call, the name was the root of the confusion since it returns true for packed YUYV/UYVY too. Added a doc note that it covers planar and packed YUV, and packed RGB stays on `GL_TEXTURE_2D`.

**Fixed the remaining "planar" wording** at the three sites: the `GlCompositor` `GL_OES_EGL_image_external`-absent warning (now "YUV … planar or packed"), the `BuildExternalProgram` comment in `gl_compositor.h`, and the `DrmCompositor` external-sampler comment in `drm_compositor.h`.

Rebuilt on the compositor config (0 errors); no `IsPlanarYuv` references remain.

jwinarske:
Addressed the latest two in d2ae62f7.

**Shared the helper.** `ExtensionSupported` now lives in `shell/backend/gl_extensions.h` as an inline `ihs::gl::ExtensionSupported`, and all three call sites use it — including `gl_caps.cc`, which had the original copy this branch was duplicating. One definition to change if the token parsing ever needs to.

**Quieted the warning.** It no longer fires from `BuildExternalProgram`, which runs on the first quad composite regardless of whether any YUV platform-view frame is ever presented — a false alarm on drivers that simply don't need the external sampler. The program is left unbuilt silently there, and a single warning is emitted only if a YUV frame actually reaches the composite guard with no sampler available.

Re-verified on `wayland-egl` with a hardware-decoded NV12 stream: zero import errors, no spurious warning (this driver has the extension), picture unchanged.

jwinarske:
Fixed in the follow-up: `gl_compositor.cc` now includes `<GLES2/gl2ext.h>` directly for `GL_TEXTURE_EXTERNAL_OES` instead of relying on the transitive include via `gl_caps.h`. (`egl_dmabuf_import.cc`, the other user of the token, already includes both `gl2.h` and `gl2ext.h`.) Rebuilt clean on the compositor config.

PR #342, opened 2026-07-24
Export the Flutter desktop C API from ihs_shared for out-of-tree plugins

jwinarske:
## What

An out-of-tree Dart FFI plugin built to the standard embedder headers must
resolve the `FlutterDesktop*` C API at load. That surface operates on engine
state and is implemented in the shell — which is an **executable**, not a
shared object a `dlopen`'d plugin can bind to. Publishing it from the
executable would make the binary its own plugin-ABI provider and split the
host contract across two artifacts.

This routes the surface through `libihs_shared.so` instead — the one library a
plugin already links for the `ihs_*` ABI — so a plugin resolves the whole host
contract from a single shared object, regardless of what process loaded it.

## How

- `ihs_shared` defines the 18 `FlutterDesktop*` functions as thin **forwarders**
  over a proc table (`include/ihs/flutter_desktop_bridge.h`,
  `src/flutter_desktop_bridge.cc`), the same shape as
  `ihs_trace_set_engine_procs`.
- The shell installs that table at engine bring-up
  (`shell/libflutter_engine.cc`), filling it with the real implementations'
  addresses — which bind unambiguously inside the executable, where
  `platform_homescreen` is linked.
- Before the table is installed the forwarders return safe defaults. The atomic
  proc-table pointer pairs a release store in the setter with an acquire load in
  each forwarder, so a plugin thread never sees a half-published table.
- `platform_homescreen` keeps its own copies for the shell's internal use, but
  they are given **hidden visibility**. Otherwise the colliding definitions in
  `ihs_shared` force the linker to export the executable's copies too, and —
  being first in symbol-search order — the executable would preempt the
  forwarders, leaving the re-export dead. Hidden, `ihs_shared` is the sole
  exporter and the executable's dynamic symbol table stays empty.
- The standalone ABI-gate build has no embedder to install the table and no
  access to the embedder headers (they live under `third_party/flutter`), so the
  forwarder source is gated on those headers being present; that build's `ihs_*`
  surface is unchanged.

## Changed from the first approach

The original revision exported `FlutterDesktop*` from the **executable** via a
version script + `ENABLE_EXPORTS`. That published the plugin ABI from the binary
rather than from the shared library plugins already link, and tripped lld on a
version-script entry that was a callback typedef, not a function. Reworked to
export from `ihs_shared` as described above; the executable now exports nothing.

## Verification (clang + lld)

| check | result |
|---|---|
| `ihs_shared` builds under lld with the forwarders | pass |
| exported symbols | 18 `FlutterDesktop*` + `ihs_*`, nothing else leaks |
| full shell builds; `libflutter_engine.cc` compiles and links `--no-undefined` | pass |
| executable `.dynsym` exports | 0 (`ihs_shared` is the sole exporter) |
| `libihs_shared.so.1` is `DT_NEEDED` of the executable (forwarders in global scope) | pass |
| runtime forwarder round-trip (default before install → forwards after install → default after clear) | pass |

### DIFF
--- a/shared/CMakeLists.txt
+++ b/shared/CMakeLists.txt
@@ -148,6 +148,30 @@ target_sources(ihs_shared PRIVATE
 )
 target_link_libraries(ihs_shared PRIVATE ${CMAKE_DL_LIBS})
 
+# Flutter desktop embedder C API forwarders (the FlutterDesktop* surface; see
+# include/ihs/flutter_desktop_bridge.h). They forward to a proc table the shell
+# installs at engine bring-up, so a plugin dlopen'd into the executable resolves
+# the whole host ABI -- ihs_* and FlutterDesktop* alike -- from this one .so.
+#
+# Only meaningful inside the homescreen tree: the standalone ABI-gate build has
+# no embedder to install the table, and the public embedder headers live under
+# third_party/flutter, outside this directory. So the source is gated on those
+# headers being present; the standalone build simply omits the forwarders (its
+# ihs_* surface is unchanged). FLUTTER_DESKTOP_LIBRARY (defined in the source)
+# makes the FLUTTER_EXPORT annotations resolve to default visibility despite the
+# hidden preset; the version script then keeps them in the dynamic table.
+set(_ihs_flutter_public
+        "${CMAKE_CURRENT_SOURCE_DIR}/../third_party/flutter/shell/platform/common/public")
+if (EXISTS "${_ihs_flutter_public}/flutter_messenger.h" AND
+        EXISTS "${_ihs_flutter_public}/flutter_plugin_registrar.h" AND
+        EXISTS "${_ihs_flutter_public}/flutter_texture_registrar.h")
+    target_sources(ihs_shared PRIVATE src/flutter_desktop_bridge.cc)
+    # Only the forwarder source includes the embedder headers, but attach the
+    # path at target scope: it applies consistently across generators and IDEs,
+    # and no other source pulls those headers in.
+    target_include_directories(ihs_shared PRIVATE "${_ihs_flutter_public}")
+endif ()
+
 #
 # Install + CMake package export
 #
@@ -159,9 +183,18 @@ install(TARGETS ihs_shared EXPORT ihs_shared_targets
 )
 
 # Public headers (hand-written) + the generated export header.
+#
+# flutter_desktop_bridge.h is deliberately not installed. Its one exported
+# symbol, ihs_flutter_desktop_set_procs, is a real ihs_* entry point (the shell
+# calls it to install the proc table), but the header that declares it #includes
+# the Flutter embedder headers, which ihs_shared does not ship -- so an installed
+# copy could not compile standalone. It is a shell-internal integration point,
+# consumed in-tree from the source include directory; nothing out-of-tree needs
+# to declare the setter.
 install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include/ihs
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
         FILES_MATCHING PATTERN "*.h"
+        PATTERN "flutter_desktop_bridge.h" EXCLUDE
 )
 install(FILES ${CMAKE_CURRENT_BINARY_DIR}/include/ihs/ihs_export.h
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/ihs
--- a/shared/ihs_shared.map
+++ b/shared/ihs_shared.map
@@ -1,12 +1,23 @@
 # Linker version script for libihs_shared.so.
 #
-# Only the sanctioned C ABI (ihs_*) is exported; every other symbol — including
-# the C++ internals of the logging bridge and any vendored dependency compiled
-# in — stays local regardless of its visibility attributes. This is the hard
-# boundary: a plugin can reach nothing but the ihs_* surface.
+# Two sanctioned surfaces are exported; every other symbol — including the C++
+# internals of the logging bridge and any vendored dependency compiled in —
+# stays local regardless of its visibility attributes. This is the hard
+# boundary: a plugin can reach nothing but these.
+#
+#   ihs_*           the native C ABI (logging, tracing, platform-view surface
+#                   negotiation, config, location).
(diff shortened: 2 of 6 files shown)


jwinarske:
Good catch — fixed. `target_link_options` is 3.13; the version script is now appended to the target's `LINK_FLAGS` via `set_property(... APPEND_STRING ...)`, which is available in 3.10. Appended rather than set, so it can't clobber a `LINK_FLAGS` applied elsewhere on the target.

Re-verified after the change, since it swapped the mechanism: 18 symbols exported, 54 dynamic in total, 0 of the plugin's 17 missing, and the real plugin still loads into the real `homescreen` process (`exit 0`).

One thing worth separating out, though: **the 3.10.2 minimum is already inaccurate on `v2.0`**, independently of this PR. Configuring with CMake 3.10 would fail today on:

- `shell/CMakeLists.txt:463` — `target_link_directories` (3.13+)
- `shared/CMakeLists.txt:79` and `:101` — `target_link_options` (3.13+)

So my line would have been the fourth such use, not the first. I've made it 3.10-safe anyway because it costs nothing and keeps this PR to its own subject — but the declaration is worth either bumping to 3.13 or fixing those three, as a separate change. Happy to do that if you'd like it, though it seemed wrong to fold a repo-wide CMake policy decision into a PR about symbol exports.

jwinarske:
CI fixed.

```
ld.lld: error: version script assignment of 'IHS_PLUGIN_ABI_1' to symbol
        'FlutterDesktopOnPluginRegistrarDestroyed' failed: symbol not defined
```

`FlutterDesktopOnPluginRegistrarDestroyed` is a **callback typedef** in the public header, not an exported function, so nothing defines it. I had added it (and one other) on the reasoning "export the whole documented C API" — read off the headers by hand rather than derived. GNU ld silently accepts a version script naming an undefined symbol; lld rejects it, so my local build passed and every clang job failed.

The list is now the intersection of *declared `FLUTTER_EXPORT` in the public headers* and *actually defined in this shell*: 18 entries, all real. `FlutterDesktopPluginRegistrarSetDestructionHandler` is defined and stays.

I also built locally with `-fuse-ld=lld` this time rather than trusting GNU ld, since it's the only one that catches this class of error. Both linkers now link clean, the binary exports the same 18, none of the plugin's 17 are missing, and the plugin still loads into the real `homescreen` process (`exit 0`).

Also from the earlier comment: `target_link_options` → `set_property(... APPEND_STRING PROPERTY LINK_FLAGS ...)` for 3.10 compatibility.

jwinarske:
Both addressed — and you're right that they resolve together.

**`LINK_DEPENDS` was missing**, and that was a real bug: an edit to `ihs_plugin_abi.map` would not have relinked, so a change to the exported set would be silently ignored on an incremental build. Verified the fix rather than assuming it — recording the binary's mtime, `touch`ing the map, rebuilding: it now relinks.

**Consistency** — followed `shared/CMakeLists.txt` exactly: `GNU|Clang` gate, `target_link_options` with the `LINKER:` prefix, `LINK_DEPENDS` on the script. That also settles the earlier `target_link_options` comment: the shared library already uses it unconditionally, so this doesn't raise the effective CMake requirement, and matching the house pattern beats a lone 3.10-safe workaround. (The declared 3.10.2 minimum is still inaccurate for the reasons in my earlier comment, but that's separate from this PR.)

Verified under both linkers, since only lld caught the last problem:

| | exported | total dynsym | shell runs | plugin loads |
|---|---|---|---|---|
| gcc | 18 | 54 | ✅ | ✅ |
| lld | 18 | 18 | ✅ | ✅ |

lld's `local: *` is stricter and leaves exactly the 18; gcc keeps some runtime symbols besides. Both start normally and both load the real flutter_webrtc plugin.

jwinarske:
Addressed the review feedback on the reworked (ihs_shared forwarder) approach:

- **struct_size not validated** (`flutter_desktop_bridge.cc`) — `ihs_flutter_desktop_set_procs` now refuses a table whose `struct_size` is smaller than `sizeof(IhsFlutterDesktopProcs)` and leaves the forwarders on their safe defaults, so a version-skewed (older shell / newer `ihs_shared`) table can't cause a call through absent trailing fields. Verified at runtime: full-size table forwards, short table is rejected to the safe default.
- **Header not self-contained** (`flutter_desktop_bridge.h`) — added `<stddef.h>`/`<stdint.h>` (matching the C-ABI spelling used by `trace.h`/`logging.h`, not the `cstddef` C++ form) so the `size_t`/`int64_t`/`uint8_t` uses don't rely on transitive includes.
- **Gate only checked one header + unquoted paths** (`shared/CMakeLists.txt`) — the forwarder source is now gated on all three embedder headers it includes (`flutter_messenger.h`, `flutter_plugin_registrar.h`, `flutter_texture_registrar.h`), and the base path plus the per-source include dir are quoted.
- **Stale PR description** — the summary has been rewritten to describe the actual approach (re-export from `libihs_shared.so` via forwarders over a shell-installed proc table), not the earlier executable-export text.

The earlier inline comments about `target_link_options`/`LINK_DEPENDS`/`ENABLE_EXPORTS` and the enumerated version script on the **executable** no longer apply: that approach was dropped. The executable now exports nothing; `libihs_shared.so` is the sole exporter of the `FlutterDesktop*` surface (and of `ihs_flutter_desktop_set_procs`), and `platform_homescreen`'s internal copies are hidden so they can't preempt the forwarders.

CI is green on the prior revision (consumer-smoke, abi-gate, unit-tests, all cross-builds); re-running for this push.

jwinarske:
Addressed the two follow-up nits on the CMake gate:

- **`set_source_files_properties(... INCLUDE_DIRECTORIES ...)`** → switched to `target_include_directories(ihs_shared PRIVATE ...)` inside the same `if()`, so the include path applies consistently across generators and IDE integrations. Only the forwarder source pulls in the embedder headers, so target scope is harmless.
- **Install-exclusion comment** → reworded to make clear the header is intentionally *not installed* (it includes Flutter embedder headers ihs_shared doesn't ship, so it couldn't compile standalone) even though its symbol `ihs_flutter_desktop_set_procs` remains an exported `ihs_*` entry point consumed by the shell in-tree.

Rebuilt clean (clang + libc++, lld); exports unchanged: 18 `FlutterDesktop*` forwarders + `ihs_flutter_desktop_set_procs`, executable dynsym empty.

PR #343, opened 2026-07-24
Add opt-in per-frame timing capture to scroll_bench

werner-scholtz:
While benchmarking scroll_bench on an embedded board I wanted to see how long each frame took to build and draw, measured in release mode. The `IVI_*_PROFILE` reports show how steadily frames reach the display, but not how long the engine spent building and drawing each one. This adds that missing half, so it seemed worth upstreaming.

- Set the `TIMINGS_FILE` dart-define to write one CSV row of FrameTiming phase timestamps per frame (vsync, build start/finish, raster start/finish, wall clock).
- Stdout gets a periodic summary and a one-time first-frame marker for startup timing.

Opt-in: with `TIMINGS_FILE` unset (the default) there is no behavior change.

Verified in release mode on Linux desktop and on a TI SK-AM62 board:
| | |
| --- | --- |
| Board | TI SK-AM62 (am62xx-evm), GP silicon |
| SoC | AM62x, quad Cortex-A53 |
| GPU | PowerVR Rogue AXE-1-16M, driver 1.15 |
| Display | WaveShare panel, 60 Hz |
| OS | TISDK 10 (scarthgap), kernel 5.10 |
| Flutter | 3.38.3, release AOT bundle |

### DIFF
--- a/test/integration/scroll_bench/README.md
+++ b/test/integration/scroll_bench/README.md
@@ -28,6 +28,7 @@ number that matters for that use case.
 | `ITEMS` | `2000` | number of rows |
 | `VELOCITY` | `600` | auto-scroll speed, logical px/s |
 | `AUTOSCROLL` | `true` | set `false` for a manual (touch-driven) run |
+| `TIMINGS_FILE` | *(empty)* | per-frame engine-timing CSV path, empty disables |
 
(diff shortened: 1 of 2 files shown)


jwinarske (approved):

jwinarske:
Thanks for the contribution @werner-scholtz !

PR #347, opened 2026-07-24
platform-views: sample planar YUV on the Vulkan importer (VkSamplerYcbcrConversion)

### LINKED ISSUE #339: platform-views: sample planar YUV on the Vulkan importer (VkSamplerYcbcrConversion)
#335 taught the EGL importer to sample planar YUV through `GL_TEXTURE_EXTERNAL_OES`. The Vulkan importer has the same gap and is not covered by that change.

`DmabufVulkanImporter::VkFormatFromFourcc` maps only the packed RGB fourccs:

```cpp
VkFormat VkFormatFromFourcc(uint32_t fourcc) {
  if (fourcc == ARGB8888 || XRGB8888) return VK_FORMAT_B8G8R8A8_UNORM;
  if (fourcc == ABGR8888 || XBGR8888) return VK_FORMAT_R8G8B8A8_UNORM;
  return VK_FORMAT_UNDEFINED;   // NV12 etc. -> rejected
}
```

So a hardware-decoded NV12 dma-buf on a Vulkan backend is refused at import ("unsupported fourcc"). On the drm-kms-vulkan backend — which is what a Pi 4 image often ships — a WebRTC video platform view therefore shows nothing (and on the EGL path before #335 it showed a red luma ramp; on Vulkan it does not import at all).

## What it needs

The Vulkan equivalent of the external-OES path:

- Map planar fourccs to their multi-planar formats: NV12 → `VK_FORMAT_G8_B8R8_2PLANE_420_UNORM`, and the others `NeedsExternalOes()` already enumerates for EGL.
- Create a `VkSamplerYcbcrConversion` (model/range/chroma-location from the frame's `color_space`/`color_range`, defaulting BT.601/709 by size as the EGL path does) and an immutable-sampler descriptor bound to it.
- Import the dma-buf with `VK_IMAGE_CREATE_DISJOINT_BIT` when planes are on separate objects (here they share one), the per-plane `VkImageDrmFormatModifierExplicitCreateInfoEXT` layout, and `VK_EXTERNAL_MEMORY_HANDLE_TYPE_DMA_BUF_BIT_EXT`.
- A `samplerExternalOES`-equivalent isn't needed — the YCbCr sampler does the conversion — but the sampling pipeline needs the immutable-sampler descriptor set layout.

This is parallel in size to #335's EGL work, not a small follow-on, which is why it is filed rather than folded in. It needs the Vulkan compositor's descriptor/pipeline setup, and hardware with a Vulkan backend to verify (the AMD desktop over drm-kms-vulkan, or the Pi's Mali/panvk stack).

Until then, planar YUV platform views require an EGL backend; that is now documented in the ihs_webrtc_view example.

jwinarske:
## What

A hardware-decoded NV12 dma-buf on a Vulkan backend was refused at import
(`VkFormatFromFourcc` mapped only packed RGB), so a WebRTC video platform view
showed nothing on `drm-kms-vulkan` / `wayland-vulkan`. The Vulkan counterpart to
the EGL gap #335 closed; implements #339.

## How

- **Importer** (`dmabuf_vulkan_import.*`): import NV12 as a non-disjoint
  two-plane `VK_FORMAT_G8_B8R8_2PLANE_420_UNORM` (planes share one dma-buf,
  offsets within the single allocation). Resolve the frame's
  `color_space`/`color_range` into a `VkSamplerYcbcrModelConversion` /
  `VkSamplerYcbcrRange`: `IHS_COLOR_SPACE_DEFAULT` → BT.601/709 by height,
  `IHS_COLOR_RANGE_DEFAULT` → studio range. First shell implementation of the
  `platform_view.h` contract the EGL path leaves to the GL driver.
- **Compositor** (`wl_layer_compositor.*`): a ycbcr conversion can only be used
  through an immutable sampler, baked into the descriptor set layout and hence a
  distinct pipeline. Build a YUV program (conversion + immutable sampler + set
  layout + pipeline) lazily, cached per `(format, model, range)` — one per video
  stream. RGB path unchanged; the two pipelines share all state and shaders and
  differ only in the layout, so the pipeline is bound per layer (viewport/scissor
  stay dynamic). The YUV image view carries the same conversion.
- **Plumbing**: the producer exposes format + colorimetry through new
  `GetVulkanImageFormat` / `GetVulkanYcbcrModel` / `GetVulkanYcbcrRange` seams on
  `ICompositorSurface`; an RGB producer returns 0 and keeps the historical
  `B8G8R8A8_UNORM` contract. The device enables `samplerYcbcrConversion`, queried
  and chained only where supported.

## Failure modes

A device without the feature (or without linear ycbcr filtering) fails
conversion creation and the layer is skipped rather than drawn wrong. Chroma
siting is not carried by `IhsFrame` and defaults to the H.264/HEVC 4:2:0
convention (cosited-x, midpoint-y).

## Verification

Builds clean on `wayland-vulkan` (clang + libc++, lld): all changed TUs compile,
the executable links, clang-tidy and clang-format clean. **On-screen
verification is pending** — needs a Vulkan backend on real hardware (AMD over
`drm-kms-vulkan`, or the Pi's Mali/panvk stack) to confirm colour and chroma
siting. EGL backends remain the verified path.

Closes #339

### DIFF
--- a/shell/backend/wayland_vulkan/wayland_vulkan.cc
+++ b/shell/backend/wayland_vulkan/wayland_vulkan.cc
@@ -781,6 +781,32 @@ void WaylandVulkanBackend::createLogicalDevice() {
     }
   }
 
+  // Enable samplerYcbcrConversion so the platform-view compositor can build a
+  // VkSamplerYcbcrConversion to sample planar YUV (NV12) video dma-bufs. Core
+  // in Vulkan 1.1 (this device targets 1.1), but the feature bit must still be
+  // requested; query it and chain it only when the device reports support, so a
+  // device without it still creates rather than failing here. Without the
+  // feature the compositor's VkSamplerYcbcrConversion/pipeline creation fails
+  // (LayerCompositor logs it once and caches the failure); a planar-YUV layer
+  // is then skipped and its video does not display.
+  VkPhysicalDeviceSamplerYcbcrConversionFeatures ycbcr_feature{};
+  ycbcr_feature.sType =
+      VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SAMPLER_YCBCR_CONVERSION_FEATURES;
+  {
+    VkPhysicalDeviceSamplerYcbcrConversionFeatures ycbcr_supported{};
+    ycbcr_supported.sType =
+        VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SAMPLER_YCBCR_CONVERSION_FEATURES;
+    VkPhysicalDeviceFeatures2 features2{};
+    features2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
+    features2.pNext = &ycbcr_supported;
+    d().vkGetPhysicalDeviceFeatures2(physical_device_, &features2);
+    if (ycbcr_supported.samplerYcbcrConversion == VK_TRUE) {
+      ycbcr_feature.samplerYcbcrConversion = VK_TRUE;
+      ycbcr_feature.pNext = const_cast<void*>(device_info.pNext);
+      device_info.pNext = &ycbcr_feature;
+    }
+  }
+
   CHECK_VK_RESULT(
       d().vkCreateDevice(physical_device_, &device_info, nullptr, &device_));
 
@@ -2617,6 +2643,10 @@ bool WaylandVulkanBackend::CompositeLayersBlend(VkCommandBuffer cmd,
     VkFormat format;
     int32_t dx, dy, dw, dh;
     VulkanBackingStore* store;  // non-null => restore to COLOR_ATTACHMENT after
+    // Color conversion for a planar-YUV source; ignored for RGB formats.
+    VkSamplerYcbcrModelConversion ycbcr_model{
+        VK_SAMPLER_YCBCR_MODEL_CONVERSION_RGB_IDENTITY};
+    VkSamplerYcbcrRange ycbcr_range{VK_SAMPLER_YCBCR_RANGE_ITU_NARROW};
   };
   std::vector<Draw> draws;
   draws.reserve(count);
@@ -2689,6 +2719,16 @@ bool WaylandVulkanBackend::CompositeLayersBlend(VkCommandBuffer cmd,
         continue;  // GL / subsurface platform view — nothing to sample
       }
       auto src = reinterpret_cast<VkImage>(vk_img);
+      // The image's real format: 0 keeps the historical B8G8R8A8_UNORM contract
+      // (RGB producers); a planar YUV producer reports its format plus the
+      // color model/range for the compositor's VkSamplerYcbcrConversion.
+      const uint32_t pv_fmt = surface->GetVulkanImageFormat();
+      const VkFormat src_format = pv_fmt != 0 ? static_cast<VkFormat>(pv_fmt)
+                                              : VK_FORMAT_B8G8R8A8_UNORM;
+      const auto ycbcr_model = static_cast<VkSamplerYcbcrModelConversion>(
+          surface->GetVulkanYcbcrModel());
+      const auto ycbcr_range =
+          static_cast<VkSamplerYcbcrRange>(surface->GetVulkanYcbcrRange());
       // Wait the producer's acquire fence (if any) in this frame's submit.
       CollectAcquireWait(surface.get());
       const auto cur =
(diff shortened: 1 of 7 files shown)


jwinarske:
Addressed all three review comments:

- **Importer plane-count / split-fd** (`dmabuf_vulkan_import.cc`) — the plane count is now derived from the format (1 for packed RGB, 2 for NV12) and the frame is rejected if `plane_count` disagrees, so an RGB fourcc can't arrive claiming two planes. Added an explicit split-fd check: `plane_fd[>0]` must be unset (-1) or the same handle as `plane_fd[0]`, else the frame is rejected rather than importing the wrong memory.
- **Ignored `vkCreateImageView` result** (`wl_layer_compositor.cc`) — `CreateSourceView` now checks the result and returns `VK_NULL_HANDLE` without recording a null view; `DrawLayer` skips the layer when the view is null instead of binding it into a descriptor set. Covers both the create site and the `DrawLayer` consumer.
- **Interface default range comment** (`compositor_surface_interface.h`) — the comment claimed "narrow" while the default returns 0 (which is `ITU_FULL`). Reworded to state the actual values (0 = `RGB_IDENTITY` model / `ITU_FULL` range), that they are only consulted for a YUV `GetVulkanImageFormat`, and that a producer reporting a YUV format must override both (there's no colour metadata to default from at the interface). The real producer already resolves both from the frame and defaults to narrow.

Rebuilt clean on `wayland-vulkan` (clang + libc++, lld); clang-tidy and clang-format clean.

jwinarske:
Addressed the second review round:

- **`GetVulkanYcbcrModel/Range` for RGB images** (`platform_view_host.cc`) — now gated on `current->yuv`: an RGB image returns the interface default (0 = RGB-identity model / ITU_FULL range) instead of the `ImportedImage` field (which defaults to narrow/601), matching the `ICompositorSurface` contract so nothing keys off a stray value on a frame never sampled through a conversion.
- **Shared-fd multi-planar contract** (`dmabuf_vulkan_import.cc`) — tightened to require `plane_fd[>0] == -1` (not merely "-1 or equal to plane_fd[0]"). Confirmed the risk: `CloseFrameFds` closes every `plane_fd[i] >= 0`, so a duplicate handle to the same numeric fd would be double-closed on a cache hit or import failure. NV12 from the VAAPI/V4L2 decoders already sets only `plane_fd[0]`.

Rebuilt clean on `wayland-vulkan`; clang-tidy and clang-format clean.

jwinarske:
Addressed the third review round:

- **Failed YUV program now cached** (`wl_layer_compositor.cc`) — a `YuvProgram` gains an `ok` flag; `GetOrCreateYuvProgram` records a negative result (null handles, `ok=false`) on any creation failure and logs it once, so a device lacking `samplerYcbcrConversion` / linear ycbcr filtering no longer reattempts the expensive conversion/sampler/layout/pipeline creation every frame — the cached lookup returns `nullptr` immediately.
- **Import diagnostics** (`dmabuf_vulkan_import.cc`) — split the conflated plane-count / `plane_fd[0]` guard into two checks so a valid plane count with a missing fd no longer reports the wrong problem.
- **Device-feature comment** (`wayland_vulkan.cc`) — reworded to the accurate failure mode: without the feature the device still creates; the compositor's conversion/pipeline creation fails later (logged + cached), and the planar-YUV layer is skipped.
- **`GetVulkanImage` doc** (`compositor_surface_interface.h`) — "blit" → "composite": the layer compositor samples the image in a shader (SHADER_READ_ONLY_OPTIMAL) and composes with alpha, not an overwriting blit.

Rebuilt clean on `wayland-vulkan`; clang-tidy and clang-format clean.

jwinarske:
Addressed the low-confidence note on `GetOrCreateYuvProgram` (validate before build): a YUV layer's model/range are now validated up front — the model must be an actual YCbCr model (601/709/2020, not the RGB-identity value a producer that reported a YUV format but left the model unset would yield) and the range one of ITU narrow/full. Invalid colorimetry logs and caches a negative result and the layer is skipped, per the "skip rather than draw wrong" contract, instead of building a conversion that would sample wrong.

Rebuilt clean on `wayland-vulkan`; clang-tidy and clang-format clean.

jwinarske:
Addressed the two findings on `ac828998`:

- **Destructor null-guards** (`wl_layer_compositor.cc`) — the YUV-program cleanup loop now guards each `vkDestroy*` with `!= VK_NULL_HANDLE`, matching the rest of the destructor, so a cached negative result (all-null handles) doesn't rely on Vulkan accepting null destroys.
- **`BeginFrame` failure** (`wayland_vulkan.cc`) — phase 2 (DrawLayer/EndFrame) now runs only when `BeginFrame` actually opened the render pass; on failure it's skipped (and `ok` set false) so no commands are recorded outside a render pass, while phase 3 still restores the backing-store layouts regardless.

Rebuilt clean on `wayland-vulkan`; clang-tidy and clang-format clean.

jwinarske:
Addressed the shared-fd contract point (`platform_view.h:470` — "one fd may back several planes"):

- **Importer** (`dmabuf_vulkan_import.cc`) — reverted the over-strict `plane_fd[>0] == -1` requirement. A secondary plane_fd is now accepted when it is either unset (-1) or a duplicate handle of `plane_fd[0]` (the shared-fd form the contract permits); only a genuinely distinct fd (the disjoint layout this path doesn't import) is rejected.
- **`CloseFrameFds`** (`platform_view_host.cc`) — made the double-close safe at the source: it now closes each *distinct* numeric fd once (skips a plane_fd that duplicates an earlier one), so a duplicated-fd multi-planar frame is no longer double-closed on a cache hit or import failure. This is correct for every caller, not just the Vulkan path.

That resolves the earlier double-close concern without the undocumented restriction. Rebuilt clean on `wayland-vulkan`; clang-tidy and clang-format clean.

(Note: the `unit-tests` CI failure on the prior push was the pre-existing `wayland_client.hpp` codegen-ordering race in the `homescreen_ut_shell` target — reproduces on the untouched base commit — not related to this change.)

Issue #338, opened 2026-07-23
platform-views: return a release fence to the producer


jwinarske:
`HostSubmit` hardcodes the release direction:

```cpp
// Release fences (compositor -> producer) are a follow-up; nothing to return.
if (out_release_fence_fd != nullptr) {
  *out_release_fence_fd = -1;
}
```

The acquire direction is wired (a producer's `sync_file` is waited on before sampling), but the return direction is not: a producer never learns when the compositor is done scanning out a submitted frame.

## Why it matters

Without it, a producer that would retire a buffer on the release fence has to fall back to a depth window — holding a fixed fraction of its pool because it cannot tell which buffers are actually free. Measured with the WebRTC presenter that drove #335: hold depth is `min(pool/3, 8)` under this implicit path, versus 0–1 when a release fence is available. On a small pool (a V4L2 CAPTURE queue of 8–10) that is the difference between comfortable and near-starvation.

The grant already advertises `sync = EXPLICIT_*`; a producer that negotiated it is entitled to a fence back and currently gets `-1`.

## What it needs

On an explicit-sync grant, the compositor creates a fence that signals when the frame is no longer being scanned out / sampled, and returns its fd through `out_release_fence_fd`; the producer owns and closes it, and waits on it before overwriting that buffer.

- DRM plane scanout: the KMS `OUT_FENCE_PTR` of the atomic commit that presented the frame.
- GL/EGL texture path: an `EGL_KHR_fence_sync` created after the draw that samples the import, exported via `EGL_ANDROID_native_fence_sync`.
- Vulkan path: a `VkSemaphore`/`VkFence` exported as a sync_file after the present that used the image.

Scope note: the import cache aliases the producer's ring rather than copying it (see jwinarske/ihs_webrtc_view#6), so this is about *timeliness* of buffer reuse, not correctness — an implicit producer that holds a depth window is safe, just conservative.

jwinarske:
Independent corroboration from a second producer — a Shadertoy platform-view plugin rendering through the GL/EGL dma-buf path — plus one finding that I think changes the scope of this issue.

## The symptom, reproduced

Requesting `IHS_PV_SYNC_EXPLICIT_PREFERRED`, the grant comes back `IHS_PV_SYNC_IMPLICIT` and `ihs_pv_submit` returns `out_release_fence_fd = -1`, on both EGL backends:

| backend | host | result |
|---|---|---|
[...]

PR #360, opened 2026-07-26
platform-views: return a release fence on submit

jwinarske:
`HostSubmit` hardcoded `*out_release_fence_fd` to `-1`. The compositor already
exports a per-frame release fence for each sampled view (`SetReleaseFenceFd`),
so hand back a `dup` of it (#338): a producer on an explicit-sync grant waits
on it before overwriting a ring buffer the compositor may still be reading,
instead of holding a conservative depth window.

`HandBackReleaseFence` does the dup under `v->mutex` in each backend path, uses
`::dup` for consistency with the other fd dups here, and logs on failure (errno)
the way the acquire-fence dup does, so an explicit-sync stall is diagnosable. It
stays `-1` until the first composite has run or if the dup fails; the producer
owns and closes the fd handed back.

Rebased onto v2.0. The companion "schedule a frame on submit" change this branch
originally carried has since landed on v2.0 (`ScheduleEngineFrame`, with a
shutdown latch), so this is now the release-fence half alone -- one file,
`platform_view_host.cc`.

Smoke-tested against the Fluorite ihs_pv producer on the wayland-vulkan backend:
zero-copy dma-buf present enabled, all four views granted an explicit-sync
(`sync=1`) dma-buf, rendering live -- the path this fence serves.

### DIFF
--- a/shell/platform/homescreen/platform_views/platform_view_host.cc
+++ b/shell/platform/homescreen/platform_views/platform_view_host.cc
@@ -875,12 +875,44 @@ int HostGrantShmFd(void* /*user_data*/,
   return v->shm_fd;
 }
 
+// Hand back a dup of the compositor's latest release fence for this view
+// (#338), so a producer on an explicit-sync grant can wait on it before reusing
+// a ring buffer. Only explicit-sync submits get a fence: ihs_pv_submit's
+// contract keeps *out_release_fence_fd at -1 on implicit paths (where release
+// rides the callback / native protocol), and a submit is implicit exactly when
+// it carries no acquire fence. Also leaves it at -1 when there is no fence yet
+// or the dup fails -- the producer then throttles conservatively rather than
+// racing the compositor. Caller holds v->mutex (release_fence_fd is read here).
+void HandBackReleaseFence(const IhsPluginView* v,
+                          int acquire_fence_fd,
+                          int* out_release_fence_fd) {
+  if (out_release_fence_fd == nullptr || acquire_fence_fd < 0 ||
+      v->release_fence_fd < 0) {
+    return;
+  }
+  const int fd = ::dup(v->release_fence_fd);
+  if (fd < 0) {
+    // Capture errno before the log call, which may clobber it, then log so an
+    // explicit-sync stall is diagnosable, matching the acquire path.
+    const int dup_errno = errno;
+    ihs::log::warn(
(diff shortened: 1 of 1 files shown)


jwinarske:
Addressed the review feedback in 8c986d10: the release-fence dup is now factored into `HandBackReleaseFence`, used by both backend paths. It switches to `::dup` for consistency with the other fd dups here and logs on failure (errno), matching the acquire-fence dup path, so a failed dup is diagnosable instead of silent. On failure the fd stays -1 and the producer throttles conservatively.

jwinarske:
Rebased onto v2.0 (force-pushed, now a single commit `59752202`).

- The **schedule-a-frame-on-submit** half this branch originally carried has since landed on v2.0 (`ScheduleEngineFrame`, with the shutdown latch), so it's dropped here — this PR is now the **release-fence** half alone.
- Copilot's earlier feedback is folded into the commit: the dup is factored into `HandBackReleaseFence` (one place instead of two), uses `::dup`, and logs on failure like the acquire-fence path.
- Smoke-tested against the Fluorite ihs_pv producer on wayland-vulkan: zero-copy dma-buf present enabled, all four views granted an explicit-sync (`sync=1`) dma-buf, rendering live.

jwinarske:
Addressed in `b11b00ee`:

- **Implicit-sync contract** — right, `ihs_pv_submit` keeps `*out_release_fence_fd` at `-1` on implicit paths (platform_view.h:497-506). `HandBackReleaseFence` now takes `acquire_fence_fd` and returns early (leaving `-1`) when it's `< 0`, so only explicit-sync submits get a fence. Both call sites pass it; `acquire_fence_fd` is still the original value there (the stash/close happen after, or on error-return paths).
- **errno clobbering** — the dup errno is now captured into a local immediately after the failed `::dup`, before the `ihs::log::warn` call, and the saved value is logged.

The remaining comment about the `ScheduleFrame` threading wording is on `ScheduleEngineFrame`, which is v2.0 code, not this PR — after the rebase this branch is the release-fence change alone and no longer touches that comment.

Rebuilt (links clean) and clang-format-19 clean.

jwinarske:
Addressed the low-confidence comment in `859aa7dc`: reworded the acquire-fence stash comment so `-1` reads as an implicit-sync submit (no acquire fence, content already ready) rather than "stalled synchronously", matching the `ihs_pv_submit` contract — and noted it's the same condition under which `HandBackReleaseFence` returns no release fence, so the two now line up.

PR #361, opened 2026-07-27
drm_kms_egl: scan ihs_pv platform views out on a KMS plane

jwinarske:
Routes an ihs_pv platform-view producer's dma-buf onto a KMS overlay plane via the drm-cxx `LayerScene`, instead of GL-compositing it. Brings up zero-copy video on a hardware plane (validated on Pi 4 with rpi-hevc-dec HEVC).

## What changed
- **Per-plane fds** — `GetDmabuf` and the scene path carry one fd per plane, so a multi-handle NV12 buffer (rpi-hevc-dec exports Y and C as *separate* dma-bufs) scans out instead of falling back to GL.
- **Per-buffer release** — `HostSubmit` hands the producer a release eventfd per submit; `OnScanoutRelease` signals it when the frame leaves scanout, and a frame superseded before scanout is released at supersede. Implicit-sync producers get a deterministic per-buffer "free" signal without an ABI change.
- **`ExternalDmaBufPool` per view** (keyed by `buffer_id`) — a rotating decoder ring imports each slot once and reuses its cached `fb_id`, replacing the single-use `ExternalDmaBufSource` that re-imported every frame and double-closed the deduped GEM handle (AddFB2 started failing once the ring wrapped).
- **Plane-topology commits** — bringing the overlay plane up/down uses a blocking `ALLOW_MODESET` commit (as the first commit does); a transient `EBUSY` (GL→scene transition racing a pending flip) is retried a bounded number of times instead of permanently latching GL fallback for the whole session.

## Notes
- Not included here: the `emb.lock` `rpi4-trixie` target entry (build-infra, machine-specific keys) — happy to add if wanted.

### DIFF
--- a/shared/include/ihs/platform_view.h
+++ b/shared/include/ihs/platform_view.h
@@ -499,13 +499,17 @@ typedef struct IhsFrame {
  * Hand the registry a produced @frame for @view. Used by all kinds
  * (TEXTURE_DMABUF_IMPORT, SOFTWARE_SHM, and the registry-driven DRM_PLANE).
  * @acquire_fence_fd signals when the plugin's production completed (-1 for
- * implicit sync); the registry takes ownership of the fd. On an explicit-sync
- * grant *out_release_fence_fd is set to a fence that fires when the buffer is
- * free to reuse — the plugin owns and closes it, and waits on it before
- * overwriting that buffer; on implicit paths it is -1 and release rides
- * IhsPvCallbacks / the native protocol. The plugin keeps whatever buffer ring
- * it wants (double, triple, N); the registry never counts them. Returns
- * IHS_PV_OK or a negative IhsPvResult.
+ * implicit sync); the registry takes ownership of the fd. *out_release_fence_fd,
+ * when the registry sets it to a non-negative fd, is a release fence that fires
+ * when the buffer is free to reuse: the plugin owns that fd and must close it
+ * (waiting on it first before overwriting the buffer). The registry may return
+ * one under any sync mode — the DRM_PLANE path returns a per-buffer release
+ * eventfd even on an implicit-sync grant, since it has no other back-channel to
+ * the producer. -1 means no fence for this submit and release rides
+ * IhsPvCallbacks / the native protocol; a plugin must therefore always close a
+ * returned fd regardless of the negotiated sync mode. The plugin keeps whatever
+ * buffer ring it wants (double, triple, N); the registry never counts them.
+ * Returns IHS_PV_OK or a negative IhsPvResult.
  */
 IHS_EXPORT int ihs_pv_submit(IhsPlatformView* view,
                              const IhsFrame* frame,
--- a/shared/include/ihs/platform_view.h
+++ b/shared/include/ihs/platform_view.h
@@ -499,17 +499,17 @@ typedef struct IhsFrame {
  * Hand the registry a produced @frame for @view. Used by all kinds
  * (TEXTURE_DMABUF_IMPORT, SOFTWARE_SHM, and the registry-driven DRM_PLANE).
  * @acquire_fence_fd signals when the plugin's production completed (-1 for
- * implicit sync); the registry takes ownership of the fd. *out_release_fence_fd,
- * when the registry sets it to a non-negative fd, is a release fence that fires
- * when the buffer is free to reuse: the plugin owns that fd and must close it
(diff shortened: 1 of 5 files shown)


jwinarske:
Addressed both Copilot review comments in c79718fd:

1. **Unused include** — removed `external_dma_buf_source.hpp` (platform views now use `ExternalDmaBufPool`; the source type has no remaining references in this file).
2. **Mid-frame PV removal vs `blocking_modeset`** — good catch. The geometry-mismatch `remove_layer` path now increments `pv_pruned` (was `pv_reused`), so `plane_topology_changed` includes it and that frame takes the blocking `ALLOW_MODESET` commit rather than a NONBLOCK one that could hit the transient EBUSY.

Smoke-tested on Pi 4 (rpi-hevc-dec HEVC → KMS plane): 0 GL-fallback latches, 0 create failures, 0 EBUSY, `composited=0` (pure plane), 0 decode errors over the full run.

jwinarske:
Second Copilot pass — both addressed in bf40983a (docs-only):

3. **Release fd on implicit sync** (`platform_view_host.cc`) — the DRM_PLANE path intentionally returns a per-buffer release eventfd even under implicit sync (it has no other back-channel to signal the producer; the old contract's "release rides `IhsPvCallbacks`" was never a shipping mechanism). Updated the `ihs_pv_submit` contract to state a release fd may be returned on **any** sync mode and the plugin **must always close** a returned fd. Our producer already does (closes it in its PollReleases/Renegotiate/dtor paths).
4. **Per-plane fd ownership** (`compositor_surface_interface.h`) — documented that `GetDmabuf` transfers ownership of **every** populated plane fd (not just `fd[0]`) and the caller must close each after import.

Compile-clean; no runtime change (doc-only).

jwinarske:
Third + fourth Copilot passes addressed (525c5c79, 168bd66b):

5. **`OnScanoutRelease` thread doc** — relaxed: it is normally the compositor thread, but scene/pool teardown can call it from another thread, so implementations must be thread-safe (ours is — mutex-guarded).
6. **`PvFdCloser` double-close** — now closes each *distinct* plane fd once (the `Dmabuf` contract permits a single-handle layout to repeat one fd). Note our `GetDmabuf` dups every plane distinctly, so this was defensive hardening, not a live double-close.
7. **`HandBackReleaseEventfd` dup leak** — on `dup()` failure the eventfd is now closed and not stored, so no fd leak / stale `release_efds` entry.
8. **`acquire_fence_fd` doc** — corrected; the scene path now imports it and wires it to the plane `IN_FENCE_FD` (CPU-wait fallback), no longer "ignored".
9. **Delivered frame lost on GL fallback** — good catch. A first-frame PV whose pool `create`/`add_layer` fails routed the (already-delivered) frame to GL fallback without releasing it → producer could block on that slot. Now released before falling back, matching the reused-path handling.

All smoke-tested on Pi 4 after each round (0 GL latches, 0 create failures, `composited=0` pure plane, 0 decode errors, full run). No runtime regression.

jwinarske:
10. **`~IhsPluginView` / `release_efds` locking** — this one is a false positive. The destructor already acquires the mutex before touching `release_efds`:

```cpp
IhsPluginView::~IhsPluginView() {
  DisposePlugin();                                  // outside the lock (join may re-enter it)
  const std::lock_guard<std::mutex> lock(mutex);    // held for the rest of the body
  ...
  for (const auto& [buffer_id, fd] : release_efds) { if (fd >= 0) close(fd); }
  release_efds.clear();
```

So the iterate/close/clear at the flagged line is already under `mutex`, consistent with `OnScanoutRelease`/`HostSubmit`. (`DisposePlugin()` is intentionally outside the lock because the producer-thread join can re-enter it.) No change needed.

jwinarske:
Recent passes:

- **`platform_view_host.cc:560` (destructor `release_efds` race)** — false positive. The destructor acquires `mutex` (the `lock_guard` right after `DisposePlugin()`), so the `release_efds` iterate/close/clear is already under the lock. No change (detailed in an earlier comment).
- **`platform_view_host.cc:490` (`SignalRelease` ignores `eventfd_write` failure)** — addressed. It should never fail for a live blocking eventfd with a small counter, but it is no longer swallowed: a failure now logs a warning naming the `buffer_id`, so a would-be silent producer stall is diagnosable. Smoke-tested on Pi 4 (0 failures, pure-plane, full run).

PR #393, opened 2026-08-01
test/multi_touch_test: scope C3 to the Wayland backends

jwinarske:
## The problem

C3 requires 24 distinct device ids, on the premise stated in its own row: "Compositor/libinput touch ids are unbounded". That premise holds on one backend family only.

| Backend | Flutter device id | Range |
| --- | --- | --- |
| `wayland-egl`, `wayland-vulkan` | the compositor's `wl_touch` point id (`display.cc`) | unbounded |
| `drm-kms-egl`, `drm-kms-vulkan` | the libinput MT slot — `pe.device = ev.slot` (`drm_seat.cc`) | 0 .. slots-1, recycled |
| `software` | the libinput seat slot, range-checked against `kMaxTouchSlots` = 16 (`software_seat.cc`) | 0 .. 15, recycled |

On drm and software the id *is* a slot, so the distinct-id count cannot exceed the controller's slot count however much the fingers churn. C3 therefore reports 10 on a 10-slot panel and fails every time — while the bug it guards against (an unbounded id indexing a fixed `surface_x[10]`) cannot occur on those backends at all, because there is no unbounded id to overflow with.

As written, C3 reads as a regression on every drm/software run. It isn't one.

## Also corrected: the injector advice

The "Real hardware" section offers the injector as the way to grow ids — "which drives the libinput path with growing tracking ids". Its kernel tracking ids never reach the app: libinput maps them onto seat slots first, and the slot is what becomes the Flutter device id on drm/software.

Measured on a Raspberry Pi 4 driving the official 800x480 DSI panel (`ft5x06`, 10 slots confirmed via `EVIOCGABS(ABS_MT_SLOT)`) on `drm-kms-egl`: the injector's churn phase advanced tracking ids to 40, and the app reported `distinct_ids_total` = 10.

## The change

Docs only. Marks the C3 column in the expectations table with a footnote, adds a section tabulating what the device id actually is per backend and why the threshold is unreachable on two of them, and fixes the injector guidance. No thresholds or code touched — `EXPECT_CHURN_IDS` stays at 24, which is correct for the backend the check was written for.

Found while running this test on the DSI panel, which is also where #392 came from.

### DIFF
--- a/test/integration/multi_touch_test/README.md
+++ b/test/integration/multi_touch_test/README.md
@@ -25,7 +25,7 @@ arena — so the raw stream reaches the checks unmodified.
 | --- | --- | --- |
 | C1 | concurrency | ≥ `EXPECT_FINGERS` (10) simultaneous contacts with that many distinct device ids |
 | C2 | legality | per-device phase machine: no down-while-down, no move/up/cancel-while-up. Catches lost transitions and the historical cancel bug (only device 0 cancelled → 9 contacts stuck down → next session violates) |
-| C3 | churn | ≥ `EXPECT_CHURN_IDS` (24) distinct device ids across the run. Compositor/libinput touch ids are unbounded; the pre-fix embedder indexed a fixed `surface_x[10]` array by id — an out-of-bounds write for id ≥ 10 |
+| C3 | churn | ≥ `EXPECT_CHURN_IDS` (24) distinct device ids across the run. The pre-fix Wayland path indexed a fixed `surface_x[kMaxTouchFinger]` array by the `wl_touch` id — an out-of-bounds write for id ≥ 10. **Wayland backends only**; see [C3 only applies to the Wayland backends](#c3-only-applies-to-the-wayland-backends) |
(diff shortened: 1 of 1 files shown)


jwinarske:
Good catch — that sentence claimed more than the paragraph supported.

The evidence I have is drm/software only: there the injector's kernel tracking ids never reach the app at all, because libinput maps them onto seat slots before the device id is assigned (measured: tracking ids driven to 40, `distinct_ids_total` = 10 on a 10-slot panel). On Wayland the injector genuinely does automate the churn — phase D taps 30 times — and whether the ids grow past 24 is the compositor's allocation behavior, which the injector cannot force either way.

Rewritten to split those two cases instead of asserting one claim over both, and the commit message carried the same overreach, so it's corrected there too.

jwinarske:
Both points from the second pass applied.

**Backend-scoped the C3 cells.** The footnote alone left the table reading `pass` for everyone, so the cells now say it directly: `Wayland: pass — drm/software: n/a`, and the pre-fix row likewise.

**Confirmed the OOB was Wayland-specific**, rather than take it on faith. The fixed array was `wl_fixed_t surface_x[kMaxTouchFinger]` in the `touch_` struct in `shell/wayland/display.h`, indexed by the `wl_touch` id. At `62ff3a46^` that was the only file under `shell/` containing it — no drm or software seat ever had one. The README and the commit message now say that, with the location, instead of the vaguer "a fixed `surface_x[10]`".

PR #394, opened 2026-08-01
logging: say when a record was truncated

jwinarske:
## The problem

Log records are capped at `IHS_LOG_TEXT_CAPACITY` (240). `ThreadRing::push` clamps to that and copies, silently — so a clipped line reads as a complete one.

That matters because the tail is where structured payloads keep the part the reader came for. Concretely: `multi_touch_test` prints its results as one JSON line from Dart. It arrives through `Engine::onLogMessageCallback`, which formats `"{}: {}"` with tag `flutter` (9 chars), leaving 230 for the message — and the JSON is longer, so every run silently dropped the five `c1`..`c5` verdict flags at the end. The line looked complete. I only noticed because I went looking for flags that were never there.

## The change

Stamp the overflow count over the last characters of a record that did not fit:

```
before: ...ABCDEFGHIJKLMNOPQRSTUVWX
after:  ...ABCDEFGHIJKLMNOPQRST...+160
```

`"...+160"` costs one more character of message than a bare ellipsis and answers "how much did I lose", which is what decides whether to go looking for the rest.

It runs only on the truncating branch, which is rare by construction, and `push` is the single choke point every producer funnels through — `emit_raw`, `emit_fmt`, and the `ihs_log` C ABI alike — so one place covers all of them.

## Why not just raise the cap

`IHS_LOG_TEXT_CAPACITY` is public ABI in `shared/include/ihs/logging.h`, documented as the truncation limit, and every producer sizes a stack buffer from it. It also sets `sizeof(RingSlot)` against `static_assert(sizeof(RingSlot) % 64 == 0)` — 20 bytes of header plus 240 of text pads to 320, five cache lines. Raising it is an ABI change plus a ring-memory increase, and it would only move the boundary rather than remove it. A record that announces its own truncation is useful at any cap.

## Verification

x86_64, a probe linking `libihs_shared` and emitting a 399-character record, A/B against the same build with the patch stashed:

| | tail of the emitted line |
| --- | --- |
| without | `…MNOPQRSTUVWXYZABCDE` — ends mid-content, no marker |
| with | `…MNOPQRSTUVWX...+160` |

399 − 239 = 160, so the count is right. A short record is byte-identical either way. Re-ran the probe after `clang-format`, which reflowed one line, so the measurement describes the committed source.

### DIFF
--- a/shared/src/logging/thread_ring.cpp
+++ b/shared/src/logging/thread_ring.cpp
@@ -2,6 +2,7 @@
 #include "thread_ring.hpp"
 
 #include <algorithm>
+#include <cstdio>
 #include <cstring>
 #include <ctime>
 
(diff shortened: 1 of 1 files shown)


jwinarske:
Good catch, and the concrete failure mode you describe is the right one.

`snprintf` returns the length it *would* have written, so with `mark[24]` a count needing 25 bytes leaves `n` at 24 while only 23 characters plus a NUL were actually written. Copying `n` bytes then reads the NUL out of `mark` and drops it into the middle of `slot.text` — a record that ends early at a sink that stops at NUL, which is worse than the silent truncation this change exists to fix.

Two changes: the buffer is now 32, wide enough for the longest `"...+%zu"` a 64-bit `size_t` can produce (4 + 20 digits + NUL), and the copy is gated on `n < sizeof(mark)` so a truncated marker is never copied. The second is unreachable given the first, and is there because the alternative if either size changes is a corrupt record rather than a compile error — the comment says so.

Re-ran the probe after the fix: 399-character record still emits 239 characters ending `...+160`, short records unchanged.

PR #395, opened 2026-08-01
test/multi_touch_test: keep the verdict flags out of the truncation

jwinarske:
## The problem

The test prints its results as one JSON line from Dart. It reaches the log through `Engine::onLogMessageCallback`, which formats `"{}: {}"` with the tag `flutter` — 9 characters of a 239-character record — leaving 230 for the message.

The encoded summary is ~307 characters. So every run silently lost its tail, and the tail was `c1_concurrency` through `c5_cancel`: exactly the five verdict flags the test exists to report.

Measured on the Pi 4 DSI run, the line ended:

```
...,"cancel_violations":0,"c1_concurrency":tru
```

Nothing about it indicated the rest was missing. I only caught it while looking for flags that had never been printed.

## The change

**Order the flags first in `_summary()`.** The clip now lands in the counters — which a reader can live without — instead of the verdicts, which are the point. With the values from the last hardware run, all five flags survive and the cut falls mid-`mean_batch`.

**Print `MTT-CHECKS` as its own line**, 51 characters:

```
MTT-CHECKS: c1=pass c2=pass c3=FAIL c4=pass c5=FAIL
MTT-SUMMARY: {"c1_concurrency":true,...}
```

Short enough to be safe by construction rather than by arithmetic. CI should scrape that rather than parse a JSON line that may have been clipped. README updated to say so.

## Scope

The cap belongs to the embedder's log record, not to this test, so this change does not try to fix it here — it arranges for the test's output to survive it. #394 makes a clipped record announce itself (`...+<n>`), which is the complementary half: this PR stops the verdict being lost, that one stops any truncation being silent.

### DIFF
--- a/test/integration/multi_touch_test/README.md
+++ b/test/integration/multi_touch_test/README.md
@@ -29,9 +29,26 @@ arena — so the raw stream reaches the checks unmodified.
 | C4 | frame batch | contacts in one hardware scan arrive with one shared timestamp (the embedder stamps the batch once). Mean move-group size during the synchronized 10-finger drag must be ≥ `EXPECT_BATCH_MEAN` (6.0). An unbatched embedder stamps each contact separately → mean ≈ 1 |
 | C5 | cancel | after a compositor cancel, every down contact receives `PointerCancel` within 500 ms. Advisory until a cancel is observed (trigger one manually via a compositor system gesture) |
 
-`MTT-SUMMARY: {json}` is printed whenever all contacts lift.
+Whenever all contacts lift, two lines are printed:
+
+```
+MTT-CHECKS: c1=pass c2=pass c3=FAIL c4=pass c5=FAIL
+MTT-SUMMARY: {json}
+```
+
 `MULTI_TOUCH_TEST: PASS|FAIL` is printed by the Self-check button, or
 automatically with `--dart-define=SELFCHECK_AFTER_S=<seconds>`.
+
+Scrape `MTT-CHECKS`, not the JSON. The embedder caps a log record at
+`IHS_LOG_TEXT_CAPACITY` (240 bytes, less the `flutter: ` tag) and the encoded
+summary is longer than that, so it does not arrive as one line — the shell
+chunks it onto a continuation record. Nothing is lost, but a reader that
+matches a single line gets only the first piece.
+
+The summary puts the five verdict flags first, so that first piece carries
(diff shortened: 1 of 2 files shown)


jwinarske:
Both fair — the first is a claim this branch doesn't back.

**README.** It described the `...+<n>` marker as though it existed here; it's #394's behavior, and merge order between the two isn't fixed. Now states that truncation is silent today and names #394 as what changes it, so the file is accurate whichever lands first.

**File header.** It still said one JSON line is printed and that CI can scrape either, which contradicted both the new output and the README's advice. Rewritten to describe both lines and to say which to scrape and why: the short lines are safe by construction, the JSON is the one that arrives clipped.

The commit message carried the same two gaps and has been amended alongside.

jwinarske:
Reworded so this reads correctly whatever lands first.

It previously said the JSON line "arrives clipped". That is true today, but #396 chunks the engine's log messages, after which the line arrives *split across records* instead — the remainder is carried, not dropped. The advice was unaffected (a scraper reading one line still cannot parse a payload that is not confined to one), but the stated reason would have been wrong.

Now it says only that the summary does not arrive as one line, and names both outcomes: dropped today, marked by #394, carried by #396. Same treatment as the earlier #394 reference — no ordering constraint between any of the three.

jwinarske:
Rebased onto v2.0 now that #394 and #396 have landed, and validated all three together on the Pi 4 DSI panel — no touch needed, the self-check timer drives it:

```
flutter: MTT-CHECKS: c1=FAIL c2=pass c3=FAIL c4=FAIL c5=pass
flutter: MTT-SUMMARY: {"c1_concurrency":false,"c2_legality":true,"c3_churn":false,"c4_frame_batch":false,"c5_cancel":true,"peak_simultaneous":0,...,"move_groups":0,"mean_batc
flutter: h":0.0,"largest_batch":0,"cancel_observed":false,"cancel_violations":0}
flutter: MULTI_TOUCH_TEST: FAIL
```

All five flags in the first record, the remainder carried on the continuation, and `MTT-CHECKS` whole on its own line. (Every check FAIL is correct here: nothing was touched.)

Also dropped the merge-order hedging from the prose. It was written when the order was unknown; now that #396 is in, the README and the file header say plainly that the summary is chunked onto a continuation record and that a one-line reader gets the first piece — which is exactly what putting the flags first arranges for.

PR #397, opened 2026-08-01
logging: rejoin a split message so it prints as one line

jwinarske:
## The problem

A log record is one ring slot. A message longer than a slot arrived as several lines, each with its own timestamp and tag, and the reader had to put it back together. For a structured payload that means no single line is parseable:

```
[I] SHEL: flutter: MTT-SUMMARY: {"c1_concurrency":false,...,"move_groups":0,"mean_batc
[I] SHEL: flutter: h":0.0,"largest_batch":0,"cancel_observed":false,"cancel_violations":0}
```

Nothing is lost there — #396 saw to that — but neither line is JSON, and the split lands mid-token.

## The change

Split below the C ABI, rejoin at the drain, so the seam never reaches a sink.

- **`ihs_log()` takes a message of any length.** Longer than a slot, it is pushed as several slots with every piece but the last carrying `kFlagContinues`. The flag rides in the byte `RingSlot` already had spare, so slot size and its cache-line sizing are untouched.
- **The drain rejoins before writing.** A sink is handed the whole message and keeps printing one prefix per line, knowing nothing about any of this — so console, file, and any future sink benefit without changes. The partial lives on the ring rather than in a local, because a drain can end mid-sequence, and it is capped so a producer that never terminates a sequence cannot grow it without bound.
- **`emit_fmt` no longer clips at its stack buffer.** This was the one that actually bit: it formatted into `char[IHS_LOG_TEXT_CAPACITY]` and truncated *upstream* of everything above, so the tail was gone before the library saw it. Over length it now formats into a `thread_local` string, kept between calls, so a caller that logs long lines allocates once and one that never does pays nothing.

The shell's newline splitting stays — a stack trace should be one record per frame. Its hard-split of over-length runs is gone, since that hands the library several complete messages and defeats the rejoin.

## DLT is deliberately excluded

Its payload buffer belongs to a libdlt we `dlopen` and explicitly do not trust with its own bounds — `libdlt_loader` brackets it with sentinel guards and self-disables on overrun. Until this rejoin, every record handed to it was bounded by the slot; after it, a record could be up to the 64 KiB assembly cap. `DltSink` clamps to the old bound, so what DLT is asked to carry does not change.

## Verification

| | result |
| --- | --- |
| x86_64 probe, 399-byte record via the C ABI | one line, one prefix, 399 bytes |
| Pi 4, `drm-kms-egl`, `multi_touch_test` summary | one 301-byte line, `json.load` parses it, 14 keys |
| before, same box | 230 bytes + a continuation, neither piece valid JSON |

Also built with `ENABLE_DLT=ON` — a config none of the earlier logging changes were compiled against.

## Relationship to what has already landed

- #396 chunked in the shell so nothing was lost. This supersedes its hard-split half while keeping its newline half, and goes further: the pieces are now rejoined rather than left for the reader.
- #394's `...+<n>` marker stays meaningful for a record that is genuinely dropped — the assembly cap, or a ring overflow — rather than split.

### DIFF
--- a/shared/src/logging/bridge.cpp
+++ b/shared/src/logging/bridge.cpp
@@ -103,8 +103,44 @@ bool DltBridge::log(const ContextHandle& ctx,
     return false;
   }
   ThreadRing& ring = RingRegistry::thread_local_ring();
-  return ring.push(ctx.index(), static_cast<std::uint8_t>(level),
-                   message.data(), message.size());
+  const auto lvl = static_cast<std::uint8_t>(level);
+
+  // A message that fits is one slot, which is the common case and stays a
+  // single push.
+  constexpr std::size_t kFit = kSlotTextCapacity - 1;
+  if (message.size() <= kFit) {
+    return ring.push(ctx.index(), lvl, message.data(), message.size());
+  }
+
+  // Longer than a slot: split it here rather than let the slot clip it, and
+  // mark every piece but the last so the drain can put it back together. Doing
+  // this below the C ABI means a caller never has to know the slot size --
+  // ihs_log() takes a message of any length and nothing is lost.
+  //
+  // All or nothing. A run whose last piece failed to enqueue would leave the
+  // drain holding a partial with no terminator, and the next unrelated record
+  // on this ring would be appended to it and printed as though it belonged --
+  // one corrupt line, plus a lost one, from a transient full ring. So take the
+  // room up front; the consumer only ever frees more, so a count read here
+  // cannot shrink under us.
+  const std::size_t pieces = (message.size() + kFit - 1) / kFit;
+  if (ring.space() < pieces) {
+    // Not enough room for the whole message. Enqueue what one slot holds and
+    // let push() mark it clipped, which is exactly the behavior this path had
+    // before it learned to split.
+    return ring.push(ctx.index(), lvl, message.data(), message.size());
+  }
+  bool ok = true;
+  std::string_view rest = message;
+  while (!rest.empty()) {
+    const std::size_t take = rest.size() < kFit ? rest.size() : kFit;
+    const bool last = take == rest.size();
+    ok = ring.push(ctx.index(), lvl, rest.data(), take,
+                   last ? 0 : kFlagContinues) &&
+         ok;
+    rest.remove_prefix(take);
+  }
+  return ok;
 }
 
 }  // namespace ihs::dlt
--- a/shared/src/logging/dlt_sink.cpp
+++ b/shared/src/logging/dlt_sink.cpp
@@ -16,12 +16,45 @@
 
 #include "dlt_sink.hpp"
 
+#include <array>
+#include <cstdio>
+#include <cstring>
+
+#include "ring_slot.hpp"
+
 #include "libdlt_loader.hpp"
 
 namespace ihs::dlt {
 
 void DltSink::write(const LogRecord& record) noexcept {
-  loader_.emit(record.dlt_ctx, static_cast<int>(record.level), record.text);
+  // A record reaching a sink can now be longer than one ring slot: the drain
+  // rejoins a message that was split across slots, so a console or file sink
+  // prints it as one line. DLT does not get that. Its payload buffer is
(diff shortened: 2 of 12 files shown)


jwinarske:
Validated on the Pi 4 DSI panel with real touch, and followed through on what this change makes obsolete.

**Touch run** — two summary lines from one session, both parsed by `json.load`, 14 keys each, with real counters:

```
MTT-CHECKS: c1=pass c2=pass c3=FAIL c4=FAIL c5=pass
peak_simultaneous 10 · distinct_ids 10 · move_groups 270 · mean_batch 3.25 · largest_batch 6
```

(c3 fails by design on drm per #393; c4 needs the injector's synchronized drag rather than fingers.)

**#394's marker can no longer fire.** With the split below the ABI, `push()` is never handed more than fits, and `bridge.cpp` is its only caller — so that truncation branch is unreachable. Rather than revert it, the guarantee moved to where a drop is now possible: the reassembly cap, which appends `...[log record capped]` instead of ending mid-content. `push()`'s marker stays as a backstop for any future direct caller.

**#395's notes went stale in the other direction.** They described a summary arriving in pieces and told CI not to parse the JSON. It now arrives whole and parses, so the README and the file header say that, and note the flags-first order is readability rather than survival. `MTT-CHECKS` is kept — a verdict with no JSON parser needed is worth having regardless.

Rebuilt and re-verified after formatting, in all three configs including `ENABLE_DLT=ON`.

jwinarske:
All three addressed — the first was a real bug in my change.

**Orphaned partial corrupting the next record.** Right, and the mechanism is exactly as described: `ok = push(...) && ok` kept going after a failure, so a run whose last piece was dropped by a full ring left the drain holding a partial with no terminator, and the next record on that ring would close it. Fixed at the producer, all-or-nothing: `ThreadRing::space()` is checked against the piece count up front, and a message that does not fit falls back to the single clipped record this path produced before it learned to split. The count is safe to act on because the consumer only ever frees more.

Probe on a nearly-full ring — 250 fillers, then a 2000-byte message needing 9 slots with ~6 free:

| | result |
| --- | --- |
| with the guard | one clipped record `YY...+1761` (2000−239=1761), sentinel present and standalone |
| without | sentinel gone entirely |

To be precise about that A/B: the guard removes the possibility of an unterminated run, so the corruption cannot arise; in this particular run the follow-up record was dropped rather than concatenated, so what I reproduced is the loss, not a concatenation.

**Cap dropping silently.** Both halves correct. The overflow could be the final chunk, in which case no later call reached the marker branch — so the marker is now appended where the clipping happens. And the marker no longer pushes the buffer past its own cap: `kMaxText` reserves room for it inside `kMaxPartial`.

**DLT clamp.** Fixed; it now carries `...+<n>` like a clipped record, so a DLT reader can tell a clamped line from one that ended there.

Worth noting from the probe: a 100 KiB message needs ~429 slots against a 256-slot ring, so it takes the clipped-record fallback and the 64 KiB assembly cap is unreachable through a single `ihs_log()` — it is a backstop, not a path.

Rebuilt and re-verified after formatting in all three configs; the Pi summary still parses, 14 keys.

jwinarske:
Both from the latest pass are in.

**Stamp came from the last piece.** A real defect, and the reasoning is right: a rejoined message belongs at the instant it was logged. `ThreadRing` now keeps `partial_ts_ns_`, set when a partial starts and cleared with it, and the drain stamps the rejoined record from that.

Probe — a 20 000-byte message logged first, then a short one 40 ms later:

```
12:42:46.418  LONG  (logged first, 20000 chars, ~84 slots rejoined)
12:42:46.458  SHORT (logged second)
```

Ordering is right and the whole message survives. In fairness that probe confirms the behavior rather than isolating the fix — the pieces are pushed microseconds apart, so the old last-piece stamp would have looked much the same here. The skew only becomes visible when a sequence straddles a drain pass, which is exactly the case the fix is for.

**The comment was stale.** Correct — `push()` does still truncate, on the fallback the all-or-nothing guard introduced: a message that will not fit the ring is handed over whole and clipped there with its own marker. My comment predated that guard. Reworded to say where each clip happens instead of claiming push() no longer does.

Rebuilt and re-verified in all three configs; Pi summary still parses, 14 keys.

PR #402, opened 2026-08-02
display: read a connector's EDID into OutputInfo

jwinarske:
First step of the output-hotplug work. Self-contained: no behavior change beyond fields being populated, and it fixes a documented match that could never fire.

## The problem

`DrmOutputProvider::EnumerateOutputs` filled `name`, extent and refresh, and left `make` / `model` / `serial` / `mm_width` / `mm_height` at their defaults — with a comment saying EDID "is not read here yet".

That is not cosmetic. `OutputInfo::serial` was therefore *always* empty, so the `edid_serial` tier of `ResolveOutput` (`output.cc:29-48`) could never match in production. A config saying

```toml
[view.output]
serial = "..."
```

silently fell through to connector-name matching. The tier is documented in `docs/source/readme.md`, has unit coverage against synthetic data, and could not work on a real machine.

## The change

The parsing already existed — `driver_probe.cc` reads the connector's EDID property blob and hands it to `drm::display::parse_edid`, whose `ConnectorInfo` carries exactly the fields `OutputInfo` was missing. Lift that helper into `display/connector_edid.{h,cc}` so both callers share one copy, and populate the provider from it.

**No new dependency.** The helper needs libdrm and drm-cxx, both already linked wherever `drm_output_provider.cc` compiles. (The `wayland_leased_drm` block only adds `lease_client.cc`, so it is unaffected.)

## Verification

Built with `-DBUILD_BACKEND_DRM_KMS_EGL=ON`, and the new code path exercised directly against a card with two heads attached — which previously reported nothing:

```
HDMI-A-1   make=LG Electronics  model=LG HDR DQHD  serial=0x000A7BA0  300x340 mm
DP-1       make=LG Electronics  model=LG HDR DQHD  serial=0x000A7BA0  600x340 mm
```

Worth noting what that shows about serials: **both panels report the same one**, along with the same make and model. So EDID makes the tier *work* without making it *sufficient* to tell two identical monitors apart — `ResolveOutput` already warns and falls through when a serial is ambiguous, which is the correct outcome here. That is the evidence behind the follow-up step adding an integrator-supplied `output_id`.

## Next

`OutputInfo::output_id` from a udev property plus the Wayland-side identity fill; then the view binding, the `IOutputListener` implementation, park/resume, and the DRM hotplug bridge.

### DIFF
--- a/shell/CMakeLists.txt
+++ b/shell/CMakeLists.txt
@@ -319,6 +319,7 @@ if (BUILD_BACKEND_DRM_KMS_EGL)
             backend/drm_kms_egl/gl_cursor.cc
             backend/gl_process_resolver.cc
             display/drm_display.cc
+            display/connector_edid.cc
             display/drm_output_provider.cc
             display/drm_device_resolver.cc
             display/drm_mode_list.cc
@@ -403,6 +404,7 @@ if (BUILD_BACKEND_DRM_KMS_VULKAN)
             $<$<BOOL:${BUILD_COMPOSITOR}>:backend/wayland_vulkan/wl_layer_compositor.cc>
             backend/drm_kms_egl/driver_probe.cc
             display/drm_display.cc
+            display/connector_edid.cc
             display/drm_output_provider.cc
             display/drm_device_resolver.cc
             display/drm_mode_list.cc
--- a/shell/backend/drm_kms_egl/driver_probe.cc
+++ b/shell/backend/drm_kms_egl/driver_probe.cc
@@ -15,6 +15,7 @@
  */
 
 #include "backend/drm_kms_egl/driver_probe.h"
+#include "display/connector_edid.h"
 #include "logging/logging.h"
 
 #include <drm_fourcc.h>
@@ -31,8 +32,6 @@
 #include <utility>
 
 #include <drm-cxx/core/format.hpp>
-#include <drm-cxx/detail/span.hpp>
-#include <drm-cxx/display/edid.hpp>
 
 #include "logging.h"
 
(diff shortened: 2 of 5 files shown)


jwinarske:
Both addressed, and the first one led to a matching fix in the other direction.

**Missing span include.** Correct — `connector_edid.cc` used `drm::span<const uint8_t>` and got the declaration only through whatever `<drm-cxx/display/edid.hpp>` happens to pull in. It compiled, but on someone else's transitive includes. It now includes `<drm-cxx/detail/span.hpp>` itself.

**Stale comment.** Also correct, and worse than stale: after the move it had become an orphan sitting directly above `}  // namespace`, describing a function no longer in the file. It moved with the function.

**Same principle in reverse.** Chasing the first one, `driver_probe.cc` still carried `<drm-cxx/detail/span.hpp>` and `<drm-cxx/display/edid.hpp>` although it uses neither `drm::span` nor `parse_edid` any more — both left with the helper. Dropped, so neither file claims a dependency it does not have.

Rebuilt and re-verified against the two attached heads after each step; identity still reads:

```
HDMI-A-1   make=LG Electronics  model=LG HDR DQHD  serial=0x000A7BA0  300x340 mm
DP-1       make=LG Electronics  model=LG HDR DQHD  serial=0x000A7BA0  600x340 mm
```

Also rebased onto current v3.0 (`8f218990`, post-#328). The earlier CI failures on this PR were environmental, not from this change: `crash-handler`, `watchdog`, `pios` and both `publish-*` jobs all died in `packagekit_dart`'s build hook with "None of systemd;elogind;basu libraries implementing sd-bus was found", while every job that exercises this code — `drm-clang`, `drm-gcc`, all six compilers, `lint-build` — passed. #403 on the same base had no failures, and #328 running later passed all of those same jobs.

PR #403, opened 2026-08-04
docs: prefer the by-path DRM device name over cardN

jwinarske:
Docs only. Comes out of the output-hotplug work: while checking how outputs can be identified across boards, `cardN` turned out not to identify the same hardware twice.

## The measurement

Two boards running this software, same driver names, different numbers:

| driver | Raspberry Pi 4 | Raspberry Pi 5 |
| --- | --- | --- |
| `vc4-drm` | `card1` | `card0` |
| `v3d` | `card0` | `card2` |
| `drm-rp1-dsi` | — | `card1` |

`/dev/dri/cardN` is assigned in probe order. So a config pinning `card1` selects the display controller on one board and the DSI controller on the other — and fails quietly, because the path opens fine and simply refers to different hardware.

`/dev/dri/by-path/<path>-card` is derived from the hardware topology instead. Any path is opened as given, so switching costs nothing: verified by running a DSI config on the Pi 4 through `by-path/platform-gpu-card`.

## The change

- Examples and the generated reference now use the by-path form; the generator metadata was updated so its output stays reproducible (re-running it reports "nothing to update").
- New README section, **Naming a card and an output**, covering the three keys and how each is stable:
  - the **card**, by path, for the reason above;
  - the **output**, by connector name — readable, matches what the kernel, `drm_info`, Weston and sway all print, and unambiguous within a card;
  - the **role**, by udev-assigned id, for a config that must span boards.

## Two traps it records

- **Connector names are not portable either.** The DSI panel is `DSI-1` on a Pi 4 and `DSI-2` on a Pi 5 — same role, different name, different card. A config pinning `DSI-1` binds on one and silently misses on the other.
- **Compositors may rename.** Weston (`drm.c:1503-1521`) and wlroots (`drm.c:1570-1577`) pass the kernel connector name through; Mutter reports `HDMI-1` where the kernel says `HDMI-A-1`. A `wl_name` written against a desktop session may not match on target.

Matching by EDID `serial` is described as unreliable rather than recommended, because it measurably is: both Pi DSI panels publish **no EDID at all**, and the two desktop monitors used for testing share one serial, make and model.

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,7 +361,7 @@ annotated all-keys file see
 | `[view.backend.drm]` | `async_flip` | `string` | `auto` | `auto\|yes\|no` | drm |
 | `[view.backend.drm]` | `compositor` | `string` | `auto` | `auto\|planes\|gl` | drm |
 | `[view.backend.drm]` | `connector` | `string` | `(rank-pick)` | `e.g. eDP-1, HDMI-A-1` | drm/sw |
-| `[view.backend.drm]` | `device` | `string` | `(rank-pick)` | `/dev/dri/cardN` | drm/sw |
+| `[view.backend.drm]` | `device` | `string` | `(rank-pick)` | `/dev/dri/by-path/<path>-card or /dev/dri/cardN` | drm/sw |
 | `[view.backend.drm]` | `explicit_sync` | `string` | `auto` | `auto\|yes\|no` | drm |
 | `[view.backend.drm]` | `input_transforms` | `array<string>` | `[]` | `array<string>` | drm/sw |
 | `[view.backend.drm]` | `mode` | `string` | `(preferred)` | `<W>x<H>@<R>` | drm/sw |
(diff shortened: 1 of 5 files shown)


jwinarske:
Both fixed, and the second was a real defect in the guidance.

**Wrong section title.** The example pointed at "Naming an output"; the section is "Naming a card and an output". Corrected.

**`output_id` does not exist.** This is the important one, and the reviewer is exactly right: the draft told readers to name the role with `output_id` and match on it, but this tree has no such key — only `wl_name`, `drm_connector`, `serial` and `index`. The advice could not be followed as written. It came from me writing up a mechanism that is planned for a later change as though it had already landed.

Rewritten to document only what exists. The section now says plainly that **no key is portable across boards** — every option is stable per board and none survives a hardware change, so a config spanning boards needs one file per board today — and notes a role-naming mechanism as planned rather than describing it as usable. The `serial` bullet keeps the measured reasons it is a last resort (no EDID on either Pi DSI panel; identical monitors sharing a serial).

Checked mechanically afterwards: every config key the section names now exists in `shell/display/output.h`, and `output_id` appears nowhere in the file. The generator still reports "nothing to update", so docs and generator remain in step.

PR #406, opened 2026-08-04
display: match an output by an integrator-assigned role name

jwinarske:
Phase 1b of the output-hotplug work, and the last identity piece before behavior changes. **DRM backends only** — see the scope note at the end.

## The problem, measured

Every key a view can be pinned with is stable for exactly one board:

| | Raspberry Pi 4 | Raspberry Pi 5 |
| --- | --- | --- |
| DSI connector | `card1-DSI-1` | `card1-DSI-2` |
| card driver | `vc4-drm` (shared with HDMI) | `drm-rp1-dsi` (its own card) |
| EDID | **0 bytes** | **0 bytes** |
| `cardN` for `vc4-drm` | `card1` | `card0` |

The panel filling the same role has a different connector name on a different card; neither publishes an EDID, so `make`/`model`/`serial` are empty and the serial tier cannot fire; and the card number moves too. **Nothing the hardware offers names the role a display plays.**

## The change

udev models DRM connectors as devices (`DEVTYPE=drm_connector`), so a rule can attach a property to one. Read it into `OutputInfo::output_id` and match on it, above the serial:

```toml
[view.output]
output_id = 'cluster'
```

Above the serial because it is assigned deliberately, works on panels with no EDID, and is the only key that survives a change of board. The order becomes `output_id > edid_serial > name > index`.

**A named role that is absent parks the view** rather than falling through. Naming a role states which physical display a view belongs on; binding an instrument cluster to whatever else is plugged in is not a graceful degradation.

**A role claimed by two connected outputs also parks**, loudly. That is a mistake in the rule, and the enumeration order that would otherwise pick a winner is not stable.

**An empty `output_id` is unset**, not a constraint nothing satisfies — otherwise a config that expanded to `""` would strand the view and blame a missing rule for it.

**Costs nothing when no rule is installed** — the map comes back empty and every existing config resolves exactly as before.

## Verified on both boards, with one config

```
Pi 4   output_id = 'cluster'  ->  [OutputManager] view bound to output 'DSI-1'
Pi 5   output_id = 'cluster'  ->  [OutputManager] view bound to output 'DSI-2'
```

Same key, right panel, different hardware.

Hardware also caught a real bug. The first version derived the card's sysfs name from the device path's basename — but the config names the device through `/dev/dri/by-path/platform-gpu-card`, which **this repo's own documentation recommends** (#403, because `cardN` moves), and `platform-gpu-card` is not a sysfs name. Every view parked. It now resolves the device number of the open fd, which handles the symlink and works for a leased fd with no path at all.

## Scope: DRM only

`DrmOutputProvider` fills the field; the Wayland `Display` does not. A `wl_output` carries no connector identity a client can join on directly — the name a compositor advertises need not be the kernel's (Mutter reports `HDMI-1` for `HDMI-A-1`, while Weston and wlroots pass it through) — so that path needs a normalization step and its own verification on a Wayland host. Until then an `output_id` match on Wayland finds nothing and parks, which the resolver reports. The resolver itself does not check the backend family and will work the day a provider populates it.

## Tests

`output_resolver-test`: 22 cases, 10 new — binding across both boards, family-agnostic matching, above the serial, parking on a miss *even when a connector name would have matched*, parking on a duplicated role, ignoring a disconnected port carrying the role, leaving the other tiers untouched with no rule installed, and constraint accounting for `empty()`.

## Follow-up

The rule itself and its documentation are Phase 6; the example is written and passes `udevadm verify` on both boards. `output_id` is deliberately **not** documented in the README here — that lands with the rule, so the docs never describe a key ahead of the code (#403 review).

### DIFF
--- a/shell/CMakeLists.txt
+++ b/shell/CMakeLists.txt
@@ -320,6 +320,7 @@ if (BUILD_BACKEND_DRM_KMS_EGL)
             backend/gl_process_resolver.cc
             display/drm_display.cc
             display/connector_edid.cc
+            display/output_identity.cc
             display/drm_output_provider.cc
             display/drm_device_resolver.cc
             display/drm_mode_list.cc
@@ -405,6 +406,7 @@ if (BUILD_BACKEND_DRM_KMS_VULKAN)
             backend/drm_kms_egl/driver_probe.cc
             display/drm_display.cc
             display/connector_edid.cc
+            display/output_identity.cc
             display/drm_output_provider.cc
             display/drm_device_resolver.cc
             display/drm_mode_list.cc
--- a/shell/configuration/configuration.cc
+++ b/shell/configuration/configuration.cc
@@ -90,6 +90,15 @@ void ParseOutputMatch(const toml::table& t, homescreen::OutputMatch& out) {
   if (t["drm_connector"].is_string()) {
     out.drm_connector = t["drm_connector"].as_string()->value_or("");
   }
+  if (t["output_id"].is_string()) {
+    // Empty is unset, not a constraint that matches nothing. Unlike the other
+    // match fields this tier parks on a miss, so an output_id that expanded to
+    // "" would strand the view and blame a missing udev rule for it.
+    if (auto v = t["output_id"].as_string()->value_or(std::string{});
+        !v.empty()) {
+      out.output_id = std::move(v);
+    }
+  }
   if (t["serial"].is_string()) {
     out.edid_serial = t["serial"].as_string()->value_or("");
   }
--- a/shell/display/drm_output_provider.cc
+++ b/shell/display/drm_output_provider.cc
@@ -17,8 +17,15 @@
 #include "display/drm_output_provider.h"
 
 #include "display/connector_edid.h"
+#include "display/output_identity.h"
+
+#include <sys/stat.h>
+#include <sys/sysmacros.h>
 
 #include <algorithm>
+#include <filesystem>
+#include <string>
+#include <system_error>
 #include <utility>
 
 #include <drm-cxx/core/device.hpp>
@@ -34,6 +41,37 @@ DrmOutputProvider::DrmOutputProvider(std::string device_path)
 void DrmOutputProvider::SetEnumerationFd(int fd) {
   enumeration_fd_ = fd;
 }
+
+namespace {
+
+// The sysfs name udev keys connectors by ("card1"), derived from the open fd
+// rather than from the path it came from.
+//
+// Not the path's basename: the device may legitimately be named through a
+// /dev/dri/by-path/ symlink -- which is what the documentation recommends,
+// precisely because cardN is assigned in probe order -- and "platform-gpu-card"
+// is not a sysfs name. Going through the device number resolves the symlink,
+// and works equally for a leased fd that has no path at all.
+//
+// Empty when the number cannot be resolved; ReadOutputIds then matches
+// connectors on any card, which is right for a single-card system.
+std::string CardSysnameFromFd(const int fd) {
+  struct stat st{};
+  if (fd < 0 || ::fstat(fd, &st) != 0 || !S_ISCHR(st.st_mode)) {
+    return {};
+  }
+  const std::string link = "/sys/dev/char/" +
+                           std::to_string(major(st.st_rdev)) + ":" +
+                           std::to_string(minor(st.st_rdev));
+  std::error_code ec;
+  const auto target = std::filesystem::read_symlink(link, ec);
+  if (ec) {
+    return {};
+  }
+  return target.filename().string();  // ".../drm/card1" -> "card1"
+}
+
+}  // namespace
 
 std::vector<OutputInfo> DrmOutputProvider::EnumerateOutputs() const {
   std::vector<OutputInfo> outputs;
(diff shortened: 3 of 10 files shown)


jwinarske:
All three addressed. The third was a real overclaim on my part.

**Wayland was claimed and is not implemented.** Correct, and worth being blunt about: `Display::ToOutputInfo` never sets `output_id`, so an `output_id` match on Wayland always parks. My test passed only because it feeds synthetic `OutputInfo` with the field already set — it covers the matching rule, not the system. I have scoped the feature to DRM in the header, renamed that test to `RoleMatchingDoesNotDependOnBackendFamily` with a note saying no provider populates it on Wayland today, and rewritten the PR description, which asserted Wayland support it does not have.

Wayland is not hard, but it needs the connector-name normalization (Mutter reports `HDMI-1` where the kernel and both other compositors say `HDMI-A-1`) plus a rule installed on a Wayland host to verify against. I would rather ship that with evidence than claim it now.

**Duplicate roles.** Fixed, and parked rather than fallen through: a role claimed by two connected outputs is a rule bug, and unlike an ambiguous serial there is no sensible fallback — the config named a role, so binding by port instead defeats the point. Logged at error with the count.

**Empty `output_id`.** Fixed at parse: empty is unset, not a constraint nothing satisfies. You are right that this tier makes it worse than for the other keys, since it would park the view and blame a missing udev rule for a config that simply expanded to "".

22 resolver tests pass, and the end-to-end binding still works on the Pi 4 after the changes.

jwinarske:
The new comment is right, and it is the sharpest one yet.

**"Parks" was not true end to end.** `ResolveOutput` returned `nullopt` for both "no constraint" and "constrained but unsatisfied", and every caller reads `nullopt` as "use the default" — `output_manager.cc:49-58` says so outright. So a missing or duplicated role still bound the view to another output, exactly contradicting what I wrote in the code comments, the tests and the PR description.

Fixed by making the distinction expressible rather than by softening the wording:

- `OutputResolution{ kUnconstrained | kBound | kUnresolved }` and `ResolveOutputDetailed`, with `OutputManager::ResolveForViewDetailed` surfacing it. Every tier reports it, not just `output_id` — a named-but-absent `drm_connector` was equally indistinguishable before.
- The existing overloads keep their signature and today's behavior, so nothing changes for current callers.
- The fallback log now names `output_id` too, and says why the plain overload still falls back.

**Scope, stated plainly:** what this PR delivers is the resolver refusing to *substitute another tier* when a role is named and absent. Making the view actually go dark is a caller decision that needs somewhere to park a view to, and belongs with the hotplug work. I renamed the test from `MissingRoleParksRatherThanFallingThrough` to `MissingRoleDoesNotFallThroughToLowerTiers` and documented that boundary in it, since the old name asserted a system behavior the test never exercised.

New case covers the distinction directly, including for the older tiers: unconstrained vs unresolved vs bound. 23 resolver tests pass, and end-to-end role binding still works on the Pi 4.

The other three comments were re-anchored copies of ones already fixed in `04758a8b` — empty-is-unset at `configuration.cc:97`, duplicate-role parking at `output.cc:66`, and the DRM-only scoping in `output.h`. I verified each against the current tree rather than assuming.

jwinarske:
Three more, two fixed and one answered.

**Cross-card role collisions were resolved by enumeration order.** Real, and the reachability is narrower than it looks: within a card a connector sysname occurs once, so `ids.emplace` can only collide on the any-card path taken when the caller cannot identify its card. There, two cards mapping the same connector name to different roles is possible and there is nothing in scope to say which card the caller is on. Silently keeping the first is the worst option, so the entry is now dropped and the connector remembered, with an error naming both values. The role then matches no output and the view parks visibly, rather than binding to whichever card udev happened to walk first. Same-valued duplicates are not a conflict and are kept.

**Empty mapped values.** Addressed at the source rather than the consumer. `ReadOutputIds` already skips a connector whose property is absent or empty, and with the change above it no longer emits ambiguous entries either, so the map's values are non-empty by construction — that postcondition is now stated in the header, which is what makes the plain assignment in `drm_output_provider.cc` correct. I did not add a downstream guard: it would be unreachable, and the stated harm does not hold in any case, since `configuration.cc` treats an empty `output_id` as unset, so `match.output_id` is never empty and an empty `info.output_id` could only fail to match.

**"requested output is not connected" was inaccurate.** Correct — that path is also taken for a duplicated role and an ambiguous serial, where nothing is disconnected. Now "no connected output satisfies the requested match".

Verified on the Pi 4, both arms:

```
[OutputIdentity] 3 connector(s) carry IHS_OUTPUT_NAME
[OutputManager] view bound to output 'DSI-1'                          # output_id = 'cluster'
[OutputResolver] output_id 'no-such-role' matches no connected output; the view is parked
[OutputManager] no connected output satisfies the requested match ... # reworded path
```

The conflict arm itself is reasoned, not measured, and the commit message says so: it needs two cards exposing the same connector name, and neither board can produce that — the Pi 5's two cards carry DSI and HDMI respectively. Making it testable would mean exposing the merge step purely for a test; happy to do that if you would rather have it covered.

Also corrected a stale count in the commit message while I was there: it claimed 8 new cases (20 total) and the suite is 10 new (23 total). clang-tidy clean, 18/18 test binaries pass.

PR #408, opened 2026-08-04
app: notice when a view's output changes

jwinarske:
Part 3 of output hotplug renegotiation. Follows #406 and #407.

## What was missing

Every piece existed and none were connected: `Display` emits add/remove/change on `wl_output.done`, `IOutputListener` had no implementer, and `SetOutputListener` was never called. This connects them.

App watches each display that reports hotplug — one listener per display, because `IOutputListener` says *what* changed but not *where*, and App can drive several. On an event it re-resolves every view on that display and compares the answer with where the view actually is (#407).

## Two real bugs, found by testing against a real compositor

I ran this under nested headless **sway**, whose outputs can be created and unplugged on demand. That turned up two defects in the emit layer that no unit test would have caught, because App is its first listener:

1. **Removals were silently dropped.** `EmitOutputDone` returned early when no listener was attached — *before* setting `announced`. Every output is bound in the `Display` constructor and every listener attaches after it, so outputs stayed marked un-announced: their next `done` was reported as an *arrival*, and their removal wasn't reported at all, since `HandleGlobalRemove` only notifies for outputs it believes were announced.

2. **The removal callback lied about the live set.** `HandleGlobalRemove` notified *before* erasing. A listener re-reads the live set to decide what a removal means — so it still found the output that had just gone away, resolved the constraint successfully, and concluded nothing had happened. The correct outcome only arrived later via an unrelated event on another output. Pure luck, and it would not have held on a single-output system.

Both are fixed here. The first run of the feature looked like it worked; it only survived scrutiny because I checked *which* output the event named.

## Design note

The comparison is a free function, `ClassifyOutputTransition`, rather than a branch inside the listener — it's the part that can be wrong, and as a pure function of `(current, re-resolved, event)` it can be exercised without a display, an engine or a compositor.

Worth review: **an unconstrained view is never "lost."** With no `[view.output]`, the backend's own pick stands and re-resolving says nothing about it, so treating an absent output as a resolution failure would park a view that is still correctly placed.

## Scope

Acting on a transition is limited to telling that view's platform views their grant is stale — per view, not `RenegotiateAll()`, since a second view on another output of the same display has not moved. The call hops to the platform thread: events arrive on the reactor thread reading the Wayland fd, and `PlatformViewRegistry` is platform-thread only.

Moving a view to another output, or parking one whose output vanished, needs a suspend path that does not exist yet (Phase 4). A view does forget its placement when its output is lost, so the same output returning reads as an arrival rather than "already there".

## Verified

Nested sway, all four transitions:

| action | result |
|---|---|
| create the named output | `has an output to bind to (on '-', wants 'HEADLESS-9')` |
| unplug it | `lost its output` |
| change its mode | `was reconfigured` |
| unrelated output changes | silent |

Pi 4: DRM unaffected — the view still binds `DSI-1`, and no watch is installed because `DrmOutputProvider` still reports no hotplug support (Phase 5).

Builds and tests pass with `BUILD_COMPOSITOR` both **on** (20 suites) and **off** (19) — platform views are gated on it, so the renegotiate call is too.

## Tests

Eight cases on the classifier, including the two easiest to get backwards: an unconstrained view is not lost when some *other* output disappears, and a reconfiguration of an output the view is not on is not an event for it.

**Not unit-tested:** the listener wiring and thread hop — those need a compositor, and are covered by the sway runs above.

### DIFF
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -32,6 +32,11 @@
 
 #include "config/common.h"
 
+#include "display/output_manager.h"
+#include "display/output_provider.h"
+#if BUILD_COMPOSITOR
+#include "platform/homescreen/platform_views/platform_view_registry.h"
+#endif
 #include "main_loop_waker.h"
 #include "timer.h"
 #include "view/flutter_view.h"
@@ -85,6 +90,45 @@ std::string ContextKey(const backend::BackendRegistry& reg,
 }
 
 }  // namespace
+
+namespace {
+
+const char* EventName(const homescreen::OutputEvent event) {
+  switch (event) {
+    case homescreen::OutputEvent::kAdded:
+      return "appeared";
+    case homescreen::OutputEvent::kRemoved:
+      return "disappeared";
+    case homescreen::OutputEvent::kChanged:
+      return "changed";
+  }
+  return "changed";
+}
+
+}  // namespace
+
+// One display's outputs, reported to App with the display it happened on.
+class App::OutputWatch final : public homescreen::IOutputListener {
+ public:
+  OutputWatch(App* app, IDisplay* display) : app_(app), display_(display) {}
+
+  void OnOutputAdded(const homescreen::OutputInfo& output) override {
+    app_->OnDisplayOutputsChanged(display_, homescreen::OutputEvent::kAdded,
+                                  output.name);
+  }
+  void OnOutputRemoved(const std::string_view name) override {
+    app_->OnDisplayOutputsChanged(display_, homescreen::OutputEvent::kRemoved,
+                                  name);
+  }
+  void OnOutputChanged(const homescreen::OutputInfo& output) override {
+    app_->OnDisplayOutputsChanged(display_, homescreen::OutputEvent::kChanged,
+                                  output.name);
+  }
+
+ private:
+  App* app_;
+  IDisplay* display_;
+};
 
 std::vector<std::shared_ptr<IDisplay>> App::BuildDisplays(
     const std::vector<Configuration::Config>& configs) {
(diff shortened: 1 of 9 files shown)


jwinarske:
Correct, and the log was overstating in exactly one direction: it claimed an action that a `BUILD_COMPOSITOR=OFF` build never takes.

Rather than gate the suffix, I moved the claim to where the action is. The transition line now states the transition and nothing more, and `RenegotiateView` reports what it actually did — which also covers a case the suffix got wrong even with the compositor **on**: a view with no platform views registered, where "renegotiating" was equally untrue.

Verified against a live mode change on nested sway, in both configurations:

```
# BUILD_COMPOSITOR=ON
[App] output 'HEADLESS-2' changed: view 0 was reconfigured (on 'HEADLESS-2', wants 'HEADLESS-2')
[App] view 0: renegotiated 0/0 platform view(s)

# BUILD_COMPOSITOR=OFF
[App] output 'HEADLESS-2' changed: view 0 was reconfigured (on 'HEADLESS-2', wants 'HEADLESS-2')
[App] view 0: platform views are not built; nothing to notify
```

The `0/0` is honest rather than a gap — the test bundle registers no platform views.

Builds and tests pass both ways (20 suites with the compositor on, 19 off); clang-tidy clean on `app.cc`.

jwinarske:
All three addressed. The third was a real functional gap, and it caught a test asserting a branch the running system could not reach.

**`<string_view>` missing from `output.h`** — confirmed: two uses, no include, working only by transitive luck. Added.

**Pre-announcing outputs that have not completed their first `done`** — correct, and the suggested set is exactly right: `SetOutputListener` now marks only what `EnumerateOutputs()` exposes (`oi->done`). A global that is bound but not yet described genuinely has not been announced, and its first `done` is an arrival. Narrow but reachable, since views are built between the `Display` constructor and the listener attaching.

**Unconstrained Wayland views could never be `kReconfigured`** — the important one, and right. `m_wayland_bound_output` was keyed on the constraint resolving, so an unconstrained view could not name its own output, `on_the_changed_output` was never true, and a mode change underneath it produced nothing. Worth stating plainly: I have a unit case asserting `kReconfigured` for exactly that shape, and it passed against synthetic input the system was incapable of producing.

Fixed as suggested — the view now records the output it landed on **by index**, so the record exists whether placement came from `[view.output]`, from `wl_output_index`, or from the default:

```
# before: "[FlutterView] view 0 has no output binding to track"
# after, no [view.output] at all:
[FlutterView] view 0 is on output 'HEADLESS-2'
[App] output 'HEADLESS-2' changed: view 0 was reconfigured (on 'HEADLESS-2', wants '-')
```

Constrained paths re-checked on nested sway for regressions, since every Wayland view now carries a binding. A view constrained to an absent output starts on the default, classifies `kLost` on the first event (clearing the stale record), then `kAppeared` when its output arrives. The subsequent unplug is deliberately silent: the view never actually moved there, and recording a binding it does not have would repeat the mistake this PR fixes. Phase 4 sets it when the view really moves.

Pi 4 re-verified: DRM unaffected, still binds `DSI-1`, clean teardown. Builds and tests pass with `BUILD_COMPOSITOR` on (20) and off (19); clang-tidy clean on the changed files.

PR #410, opened 2026-08-04
view: park a view whose output went away

jwinarske:
Second half of Phase 4. Follows #409.

Phase 3 detected that a view had lost its output and did nothing but notify. This applies the `[view.output]` `on_disconnect` policy: park on loss, unpark when the output returns.

**Updated:** an earlier revision of this PR said it could not stop the view's own rendering, on the grounds that `StopVsyncMonitor` is irreversible in three backends. That was the wrong lever — the right one is the vsync baton, and @jwinarske pointed it out. It now stops frame production too.

## What parking does

- **Plugins stop producing** — `set_suspended(1)` reaches the view's platform views (#409). That is where most of the cost of an unseen view sits: a camera or video plugin holding a decoder busy for a display nobody can look at.
- **The view stops rendering** — by withholding the vsync baton. Flutter asks for a frame and gets no answer, so it builds none and the UI thread, rasterizer, GPU and scanout all go quiet.

Withholding the baton is what `IVsyncProvider` *already* does during bring-up: a baton arriving before the task runner is wired is left parked until `SetRunner` drains it. This is that mechanism given a second trigger, not a new one — and it lives in the one class every vsync-driving backend funnels through, so a single gate covers all seven instead of each growing its own pause. It is reversible, which `StopVsyncMonitor` is not.

Engine and surface are untouched, so returning is immediate.

## Still not implemented, and said so in code

`on_disconnect = teardown` needs the view destroyed and rebuilt, which nothing does yet. It warns and parks instead rather than appearing to honor the config. `preload` is untouched for the same reason.

## On the review comments

The two `Suspend()`/`Resume()` ordering comments were right about the bug — a failed post could permanently suppress notifications. I did not fix them the way suggested, because deferring the state flip now has a worse failure: the vsync gate would be parked while `m_suspended` stayed false, so a later `Resume()` would early-return and **never hand the baton back**, leaving the view frozen. Instead both paths go through one `SetSuspended(bool)`: the state tracks what the vsync gate was told, and a failed post is logged. A post can only fail when no engine is running — in which case no plugin can have registered either, so there is nobody to notify.

The stale `RenegotiateView` comment is corrected.

## Tests

Four cases on the gate (21 suites pass, 20 with the compositor off). They need no engine: `PostOnVsync` no-ops without one, and `HasParkedBaton()` reports whether the engine is still waiting.

- a delivery while parked leaves the baton held — this is the mechanism
- a *discard* does too (that path exists to keep Flutter scheduling, the opposite of parking)
- unparking hands the baton back
- parking an idle pipeline holds nothing

**Disabling the gate fails exactly the two that assert the mechanism**, so they are not vacuous — I checked.

## Verified

Nested sway: unplugging the bound output parks the view. Pi 4: DRM binds `DSI-1` as before, clean teardown.

I could not demonstrate a sustained frame-rate drop end-to-end — none of the bundles available here animate continuously in these harnesses (the `mtt` bundle renders on input; under headless sway nothing composites). The unit tests above are what actually pin the behavior.


jwinarske:
All three addressed, and the PR is reworked on your vsync point — thank you, that was the right lever and I had been looking at the wrong one.

**The baton is the pause.** I had concluded frame production could not be stopped because `StopVsyncMonitor` is irreversible in three backends. Wrong layer: withholding the baton parks the pipeline, and `IVsyncProvider` already does exactly that during bring-up (a baton arriving before the runner is wired is left parked until `SetRunner` drains it). So this is the existing mechanism given a second trigger, in the one class all seven vsync-driving backends funnel through — one gate, reversible, engine stays warm. `SetParked(true/false)` on the provider, a `Backend::SetVsyncParked` virtual, and `FlutterView::Suspend()/Resume()` drive it.

**Suspend/Resume ordering** — right about the bug, but I did not take the suggested fix, and want to be explicit about why. Now that `Suspend()` also parks vsync, deferring the state flip until the post succeeds has a worse failure mode: the gate would be parked while `m_suspended` stayed false, so a later `Resume()` would early-return and never hand the baton back — a permanently frozen view. Both paths now go through one `SetSuspended(bool)` where the state tracks what the vsync gate was told, and a failed post is logged instead of silently swallowed. A post can only fail when no engine is running, and no plugin can have registered without one, so there is genuinely nobody to notify in that case.

**The stale `RenegotiateView` comment** is corrected — it now says what is actually missing (moving a view to a different output), not parking.

Four unit cases pin the gate, and they need no engine (`PostOnVsync` no-ops without one; `HasParkedBaton()` reports whether the engine is still waiting): a delivery while parked leaves the baton held, a discard does too, unparking hands it back, and parking an idle pipeline holds nothing. I disabled the gate to check they fail — exactly the two that assert the mechanism do.

One thing I could not show end-to-end: a sustained frame-rate drop. None of the bundles available here animate continuously in these harnesses, so the unit tests are what actually pin the behavior rather than a runtime measurement. Flagging that rather than implying more coverage than there is.

Pi 4 re-checked (DRM binds `DSI-1`, clean teardown); 21 suites pass with the compositor on, 20 off; clang-tidy clean.

jwinarske:
Both new findings were real, and the first one corrects something I got wrong in my own reporting.

**`DrainParkedBaton()` bypassed the gate.** Exactly right, and it is not a corner: `SubmitBaton` takes that inline kick whenever no present is in flight, which is most of the traffic. So a fresh request from the engine was served as though nothing were parked, and parking only ever caught the source-event path.

This also explains something I misattributed. In the previous round I reported that I could not demonstrate a frame-rate drop end-to-end and blamed the test bundles for not animating. That was the wrong conclusion from the right observation — the gate genuinely was not holding, because the path carrying the requests went around it. My unit tests passed because they exercised `DeliverVsync` only. Gated now, with a case for it.

**`DeliverParkedBaton()` could lose the baton.** Also right, and it is the worse failure of the two: `PostOnVsync` silently no-ops without a runner, so exchanging first drops the baton and leaves Flutter waiting for a vsync that can never arrive. It now checks the runner first, as `DrainParkedBaton` already did, and leaves the baton parked when there is nowhere to post — recoverable, since the kick latch drains it once the runner is wired. That fixes the same latent hazard for the existing callers (VT resume, post-modeset drain), not just the unpark path.

Tests went from four to six, and the two additions are the ones that matter:

- `AFreshRequestWhileParkedIsNotServed` — the leak above
- `UnparkingWithNoRunnerKeepsTheBaton` — the loss above

They needed a real `TaskRunner` to be meaningful: with no runner wired, both the gated and ungated paths short-circuit for the same reason, so the old tests could not have caught this. The runner is wired with a null engine, which `PostOnVsync` bails on before posting, so what the assertions read is whether the baton was taken out of `vsync_baton_`.

I disabled each fix in turn to confirm the tests fail without it: removing the drain gate fails three cases, removing the runner check fails exactly the one that guards against loss.

The other three comments were re-anchored copies of ones already fixed in `8a3edd19` — `ISurfaceSink::SetVsyncParked` is declared, the stale `RenegotiateView` comment is gone, and `Suspend()`/`Resume()` both go through one `SetSuspended(bool)`. Verified against the tree rather than assumed.

Built with every backend I touched this time, including software and both Vulkan backends — the configuration gap that broke CI last push. 21 suites pass in all three local configs; Pi 4 and nested sway re-checked.

PR #411, opened 2026-08-04
drm-kms-egl: HDR — drive connector HDR_OUTPUT_METADATA from a PV's HDR frame

jwinarske:
## What

Drive the DRM connector's `HDR_OUTPUT_METADATA` from an HDR (PQ/HLG) video platform view, so a direct-scanned-out HDR video signals HDR to the sink. Two commits:

1. **`HDR_OUTPUT_METADATA` signaling** — the producer's `IhsHdrMetadata` is copied into the view at submit (the plugin pointer is nulled, never retained) and exposed as *persisted* state via `ICompositorSurface::GetHdrMetadata`. Persisted rather than per-frame so the connector signal holds steady across reuse presents; toggling it re-syncs the sink's HDR mode every few frames (flicker). In the DRM scene path the topmost HDR PV's metadata is mapped to drm-cxx's `HdrSourceMetadata` (0.00002-unit primaries → CIE xy, luminance clamped to the CTA-861.3 `uint16` fields) and applied via `LayerScene::set_output_metadata` before commit; SDR clears it. The HDR layer also carries BT.2020 `color_primaries` + its source EOTF so the scene auto-derives the connector Colorspace. No-op on connectors without `HDR_OUTPUT_METADATA`.

2. **`replace_source` PV-resize robustness** (general, split out) — when a live PV's dma-buf geometry changes under its fixed-generation pool (ABR rendition switch, or subpixel resize jitter), swap the source in place via `LayerScene::replace_source` with a fresh pool instead of prune + re-add next present. The layer keeps its plane assignment, geometry, and auto-derived HDR/Colorspace signaling → no gap frame (which blinks the view and toggles the connector's HDR/Colorspace) and no per-jitter modeset. Prune + re-add remains the fallback.

## ABI note

`Dmabuf::HdrMetadata` is a nested **type** only — there is no `Dmabuf` member of it, so the `Dmabuf` layout is unchanged. The compositor reads HDR through the new `GetHdrMetadata` virtual, not the `Dmabuf`.

## Validation

- Monitor-validated on a **Raspberry Pi 5 (HDMI, vc4)** into an HDR display: the panel enters HDR mode. The `HDR_OUTPUT_METADATA` InfoFrame this path emits (`eotf=02`/PQ, mastering 1000 nits, MaxCLL 1000/MaxFALL 400) matches byte-for-byte the drm-cxx reference `hdr_demo`, which drives the same display into HDR.
- Full host clang build (all backends + compositor): 178/178, links clean.
- clang-tidy-19 clean, clang-format-19 applied.

## Known gaps (tracked separately, not this PR)

- Connector `Colorspace` stays `Default` on vc4 even for the reference `hdr_demo` — a drm-cxx/vc4 apply issue, not needed to enter HDR.
- `max bpc` stays 8; true 10-bit needs a drm-cxx setter + a bandwidth-viable mode (4K60 10-bit RGB exceeds HDMI 2.0).

Exercised via the companion plugin PR (`LP_PV_HDR` test producer).

### DIFF
--- a/shell/backend/drm_kms_egl/drm_compositor.cc
+++ b/shell/backend/drm_kms_egl/drm_compositor.cc
@@ -2676,6 +2676,28 @@ bool DrmCompositor::PresentLayersViaScene(const FlutterLayer** layers,
         drm::span<const drm::scene::ExternalPlaneInfo>(planes.data(), np),
         std::move(acquire));
   };
+  // Stand up a fresh ExternalDmaBufPool for a PV's decoder ring at the frame's
+  // geometry. Used both on first sight and when a live PV's buffer geometry
+  // changes (swapped in place via replace_source). The deferred on_release
+  // holds a displaced buffer until the next present's flip completes, then
+  // returns it to the producer (see deferred_releases_).
+  const auto make_pv_pool = [this](
+                                const ICompositorSurface::Dmabuf& db,
+                                std::shared_ptr<ICompositorSurface> surface) {
+    drm::scene::ExternalDmaBufPool::Options opts{};
+    opts.on_release = [this, surface = std::move(surface)](
+                          std::uintptr_t key,
+                          std::optional<drm::sync::SyncFence>) {
+      if (!surface) {
+        return;
+      }
+      const std::scoped_lock lock(deferred_releases_mu_);
+      deferred_releases_.push_back({surface, static_cast<uint32_t>(key)});
+    };
+    return drm::scene::ExternalDmaBufPool::create(backend_->device(), db.width,
+                                                  db.height, db.fourcc,
+                                                  db.modifier, std::move(opts));
+  };
   for (auto& fl : frame_layers) {
     ++z_index;
     const bool is_ui = (z_index == 0);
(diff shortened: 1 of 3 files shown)


jwinarske:
Both review findings addressed in `33f71e08`:

1. **`GetHdrMetadata` compiled outside `#if IVI_HAVE_EGL`** — real build break for Vulkan-only builds (`IVI_HAVE_EGL=0`), since the override dereferences the EGL-only `pending_egl`. The override is now guarded; Vulkan-only builds fall back to the base `ICompositorSurface::GetHdrMetadata` default (`false`).

2. **`output_hdr` decided during the pre-scan** — a PV can still be pruned mid-frame (the `replace_source` fallback), which left HDR signaled over SDR content for that present. `output_hdr` is now set only from PVs that actually survive into the committed scene (reused or newly added); the add/update loop walks bottom-to-top, so the topmost surviving HDR PV wins, and a pruned PV never contributes.

Full host build (all backends + compositor) links clean; clang-tidy-19/clang-format-19 clean.

PR #412, opened 2026-08-04
drm: report connector hotplug to the output layer

jwinarske:
Phase 5 of output hotplug renegotiation. Follows #410.

The udev monitor, `IOutputListener`, and App's handler all already existed — nothing joined them, and `DrmOutputProvider::SupportsHotplug()` returned `false` to say so. A display unplugged on a DRM target went unnoticed. This joins them and flips that to the truth.

## Ownership: the display, not the session

`HotplugMonitor` is a plain udev netlink socket and needs no seat. `DrmSession` **is** absent on the direct-open path (`--drm-no-seat`) that headless and kiosk targets use — exactly where hotplug still has to work.

I hit this rather than reasoned it. My first cut hung the handler off the session, and the Pi 4 said:

```
[DrmDisplay] no session; connector hotplug will not be reported
```

Owned by the display, it comes up either way:

```
[DrmDisplay] connector hotplug wired to the App reactor
[App] watching outputs on a display (3 live now)
```

## Threading

The fd is armed on the App reactor, so the callback lands on the thread `IOutputListener` is contracted to run on — App's handler walks its views and calls into backends, and must not race the loop. `dispatch()` drains every pending uevent per readable edge, so a burst collapses into one rescan rather than one per event.

The kernel's `CONNECTOR=` hint is deliberately ignored: it's absent on older kernels and on blanket events (GPU reset), and a rescan is correct in every case — including the ones the hint would get wrong.

## The diff is where the DRM shape bites

Split out as a pure function, because this is easy to get wrong: **a connector is not created and destroyed as displays come and go.** It stays enumerated with `connected` moving, so an unplug is a flag transition — a set difference on names would report nothing at all. A connector object that *does* vanish (MST teardown, card removal) is a removal too. Ports that were never connected aren't events; every card enumerates some.

**Removals are ordered before additions.** A panel swapped on one port is a removal and an addition carrying the same name, and a listener seeing the addition first would find the view already bound and conclude nothing happened — silently breaking scenario 3.

Eight unit cases cover it, including the swap ordering and the never-connected ports.

## Verified, and what is not

On the Pi 4: the monitor opens on the **no-seat** path, arms on the reactor, App watches all three connectors, and teardown stays clean — the fd belongs to the monitor, so the descriptor `release()`s rather than closes it.

**Not verified on hardware: an actual connector transition end to end.** Both HDMI ports on that board are empty and forcing a connector state needs root. The diff is covered by the unit cases; the uevent→rescan path is wired and traced (`[DrmOutputProvider] uevent: no output changes`) but has not been fired by a real plug event. Flagging that rather than implying it.

Builds and tests pass in all three local configs including every backend touched, plus aarch64; clang-tidy clean on the changed files (the two warnings in `drm_display.cc` are pre-existing, at lines well above this diff).

### DIFF
--- a/shell/app.cc
+++ b/shell/app.cc
@@ -32,6 +32,9 @@
 
 #include "config/common.h"
 
+#if BUILD_BACKEND_DRM_KMS_EGL || BUILD_BACKEND_DRM_KMS_VULKAN
+#include "display/drm_display.h"
+#endif
 #include "display/output_manager.h"
 #include "display/output_provider.h"
 #if BUILD_COMPOSITOR
@@ -252,6 +255,17 @@ App::App(const std::vector<Configuration::Config>& configs) {
   // primary_ioc_.
   for (const auto& display : m_displays) {
     if (auto* d = dynamic_cast<Display*>(display.get())) {
+      d->SetEventLoop(primary_ioc_);
+    }
+  }
+#endif
+
+#if BUILD_BACKEND_DRM_KMS_EGL || BUILD_BACKEND_DRM_KMS_VULKAN
+  // Same for the DRM displays: their hotplug uevents arrive on the session's
+  // own thread, and installing the reactor here is what lets them be
+  // marshalled onto it before any listener is attached below.
+  for (const auto& display : m_displays) {
+    if (auto* d = dynamic_cast<DrmDisplay*>(display.get())) {
       d->SetEventLoop(primary_ioc_);
     }
   }
--- a/shell/display/drm_display.cc
+++ b/shell/display/drm_display.cc
@@ -278,6 +278,18 @@ DrmDisplay::DrmDisplay(AdoptFd,
 }
 
 DrmDisplay::~DrmDisplay() {
+  // Detach the hotplug descriptor before anything else. The fd belongs to the
+  // HotplugMonitor, so release() it -- closing it here would leave the monitor
+  // holding a stale fd, and asio would close it a second time. Cancel first so
+  // no handler is left queued on a reactor that is going away.
+  if (hotplug_fd_.has_value()) {
+    std::error_code ec;
+    hotplug_fd_->cancel(ec);
+    (void)hotplug_fd_->release();
+    hotplug_fd_.reset();
+  }
+  hotplug_.reset();
+
   // Stop the per-card flip reader first -- it runs its own thread and holds the
   // fd via flip_descriptor_, so it must be torn down before the fd is closed.
   StopFlipReader();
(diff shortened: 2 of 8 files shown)


jwinarske:
All eight addressed. Five were comments of mine that did not match the code, one was a test whose name promised behavior it did not assert, and one was a real behavior bug.

**The rescan was not coalesced** (suppressed comment, and the substantive one). Right: `dispatch()` calls the handler once per accepted uevent, and a plug event routinely produces several, so calling `RescanAndNotify()` from inside the handler re-enumerated the card per event and would report the same change repeatedly. The handler now only raises a flag and the rescan runs once after the drain — which is what the comment claimed all along.

**`event_loop_` was dead state** and **`asio/post.hpp` unused** — both true, and both leftovers from the first cut that marshalled through `DrmSession`. Once the display owned the monitor and armed its fd directly on the reactor, neither was needed. Removed, along with the duplicate `stream_descriptor.hpp` include.

**The inaccurate comments** — all correct, and worth naming as a group since they were mine:

- `OutputChange::output` now says plainly that `kRemoved` carries the *last seen* state, and that only the name is meaningful there.
- The join key is the name alone; the comment claimed `name/output_id`, which `FindByName` does not do.
- The baseline-seeding comment asserted something about the caller that is not guaranteed; it now describes the intent instead.
- The ordering rationale was wrong on its own example. A panel swap on one port is *not* a remove/add pair — the connector never reports disconnected in a single rescan — so it is one reconfiguration. The ordering matters for the genuinely mixed case (a card losing DSI as HDMI arrives), and the comment now uses that.

**The test name** followed from that same mistake: I wrote `SwappingPanelsOnOnePortRemovesBeforeItAdds`, found the implementation returned a single `kChanged`, corrected the assertion, and left the name. Renamed to `SwappingPanelsOnOnePortIsAReconfiguration`. Ordering is still covered, by the case that actually exercises it.

On the CI failure: `lint-tidy (1/4)` failed with **no failed step** — the `clang-tidy` step shows `conclusion: null` after a ~49-minute job, so it was killed rather than reporting a diagnostic. I reproduced the exact shard locally with the same `clang-tidy-19` and the same CMake configuration as `lint-build` (no `BUILD_UNIT_TESTS`), and it passes. Most likely a ctcache miss storm: touching `display/output.h` invalidates every TU that includes it, so that shard rebuilds cold. I have re-run the failed jobs.

Rebuilt in every configuration including the CI lint one, 21 suites pass, aarch64 clean, and the Pi 4 re-checked after the dispatch change: monitor opens on the no-seat path, App watches all three connectors, teardown clean.

PR #413, opened 2026-08-05
docs: document output_id, the udev rule, and what happens on unplug

jwinarske:
The last of the output hotplug series. Follows #412.

## The docs said the feature was still planned

That section ended with:

> **There is currently no key that is portable across boards.** … A role-naming mechanism (an integrator-supplied udev property) is planned to close this; until it lands, prefer per-board configuration.

`output_id` landed in #406. It now documents the mechanism, ships a worked rule, and describes what a view does when its output disappears.

## The generator's safety net had a hole

`gen_config_reference.py` promises that *"adding a key to the parser forces a doc update"* — it errors on any parsed key with no META entry. But `[view.output]` keys are read through a table helper the `at_path()` scraper cannot see, so they were carried as a **hand-maintained list**. That list silently stopped covering the table the moment someone added a key, which is precisely how `output_id` shipped undocumented.

It now scrapes `ParseOutputMatch` for `t["<field>"].is_<type>()`. Applying that immediately errored with:

```
error: parser keys with no META entry (update scripts/gen_config_reference.py):
  output.output_id
```

which is the safety net doing its job for the first time on this table.

## Two descriptions were promising behavior that does not exist

Corrected rather than left flattering:

- `on_disconnect = teardown` — warns and suspends instead of freeing the engine.
- `preload` — parsed but not acted on; a view is built either way.

The new prose also states that moving a running view to a *different* output is not implemented: the transition is detected, logged, and plugins are told their grant is stale, but the view is not rebound.

## The rule

[`docs/config-examples/99-ihs-outputs.rules`](https://github.com/toyota-connected/ivi-homescreen/blob/jw/output-docs/docs/config-examples/99-ihs-outputs.rules) covers both verification boards in one file — only the rules whose `ID_PATH` matches the running board take effect. It records the two things that cost time to work out:

- `DEVTYPE` is a property, so it must be matched as `ENV{DEVTYPE}=="drm_connector"`; `udevadm verify` rejects the bare form.
- `ID_PATH` is per **GPU**, not per connector — on a Pi 4 all three connectors share `platform-gpu` — so it must be paired with a `KERNEL` pattern.

## Checked, not remembered

Given how this series has gone, I verified each claim in the new prose against the code rather than from memory: the tier order (`output_id` above `serial`), that a miss parks, the exact parked log line, that the Wayland provider does not populate `output_id`, that `teardown` warns, that `preload` is unread anywhere, and that no rebind path exists.

Also: `udevadm verify` accepts the shipped rule, the generator is idempotent (running it twice reports "nothing to update"), and the links resolve **from the repository root** — `README.md` is the real file and `docs/source/readme.md` is a symlink to it, so my first attempt at a docs-relative path would have been broken on GitHub.

### DIFF
--- a/README.md
+++ b/README.md
@@ -380,6 +380,7 @@ annotated all-keys file see
 | `[view.output]` | `index` | `int` | `(none)` | `int` | all |
 | `[view.output]` | `name` | `string` | `(primary)` | `e.g. DP-1, HDMI-A-1` | all |
 | `[view.output]` | `on_disconnect` | `string` | `suspend` | `suspend\|teardown` | all |
+| `[view.output]` | `output_id` | `string` | `(none)` | `role name from a udev rule` | drm |
 | `[view.output]` | `preload` | `bool` | `false` | `true\|false` | all |
 | `[view.output]` | `serial` | `string` | `(none)` | `EDID serial` | drm |
 | `[view.output]` | `touch_device` | `string` | `(primary)` | `device-name substring` | all |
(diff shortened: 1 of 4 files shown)


jwinarske:
Right on both counts, and the second one matters more than it looks: my scraper was incomplete in the same way the hand-maintained list it replaced was. Recognizing only `is_string`/`is_integer`/`is_boolean` would have put the hole straight back for the first `[view.output]` key parsed with a different predicate — having just fixed that exact failure, I reintroduced a narrower version of it.

Both fixed as suggested. The body is now located by a regex anchored on the definition line and the next brace at column 0, and every `is_*()` predicate counts, mapped through the existing `TYPE` table with a fallback to the raw token. I had also hand-rolled a three-entry dict when `TYPE` already covered six, including `is_floating_point` and `is_array`.

One thing I added beyond the suggestion, from the same reasoning: **a helper that no longer matches is now a hard error, not an empty result.** If the anchor silently yielded nothing, the whole table would drop out of the reference *and* take the "every parsed key needs a META entry" check with it — failing open in exactly the way this change exists to prevent.

Falsified all three, since a guarantee that has never fired is not a guarantee:

```
# a key parsed with an unlisted predicate
error: parser keys with no META entry: output.fake_gamma        (is_floating_point)
error: parser keys with no META entry: output.fake_modes        (is_array)

# the helper renamed
error: ParseOutputMatch definition not found in .../configuration.cc;
       update the scraper in gen_config_reference.py
```

and confirmed the generator still reports "nothing to update" on the real tree, with all 13 `[view.output]` keys present and no source drift.

PR #414, opened 2026-08-07
accessibility: enrich semantics nodes, roll embedder.h forward, add the hub ABI

jwinarske:
Three commits. The first retains fields the semantics mirror was dropping; the second rolls the vendored `embedder.h` forward; the third fixes a bounds bug the version audit turned up.

## Deployment context (read this first)

meta-flutter pins `FLUTTER_SDK_TAG ??= "3.38.3"`, so **3.38.3 is what we link at runtime**, while the header vendored here is **3.44.2**. The gap is narrow but real, and it drives the design below:

| Capability | 3.38.3 (deployed) | 3.44.2 (vendored header) |
|---|---|---|
| `FlutterSemanticsFlags` + `flags2` | ✅ | ✅ |
| `FlutterEngineSendSemanticsAction` (with `view_id`) | ✅ | ✅ |
| `ScrollToOffset` / `Expand` / `Collapse` | ✅ | ✅ |
| `FlutterSemanticsNode2.identifier` | ❌ **absent** | ✅ |
| `FlutterSemanticsFlags.is_accessibility_focus_blocked` | ❌ **absent** | ✅ |

Both deltas are **clean appends** — every shared member is at an identical offset in both revisions, verified member-by-member against `git show 3.38.3:...embedder.h`. So `struct_size` gating is exactly sufficient and nothing is silently reinterpreted.

**Practical upshot: `GetIdentifier()` returns `""` on every node on a shipping target today.** The field and its code path are live; 3.38.3 simply never writes it. That is expected, not a bug, and it resolves when the engine floor moves.

## 1. Retain actions, transform, and custom actions

A consumer of the tree could describe a node but not what it accepts.

- **`actions` bitmask** — `GetActions()` plus a `HasAction()` bit test.
- **`transform`** — parent-local bounds are not comparable across the tree without it.
- **`custom_accessibility_actions`** — the `FlutterSemanticsCustomAction2` batch was previously iterated and discarded. It now resolves into a tree-level table keyed by ID, via `FindCustomAction()`.

**Custom action declarations are deliberately not pruned with the nodes.** Flutter assigns a custom action its ID once and may declare it in an earlier batch than the node referencing it, so dropping an unreferenced entry would lose it permanently. The set is bounded by the app's distinct `CustomSemanticsAction` values, not by tree size.

Role derivation moved onto `accessibility::Translate()`, which was already written and unit-tested but had no call sites.

## 2. Roll `embedder.h` forward

`identifier` is retained and exposed. Role derivation now takes `FlutterSemanticsFlags`, falling back to the deprecated enum when `flags2` is null. The struct is the richer source: tristates distinguish "the property does not apply" from "it applies and is false", which the enum could only encode as paired bits. Both representations normalize into one internal view so the role precedence chain stays written once.

Fallout, both caught at compile time: the node's `flags` member is now `flags__deprecated__`; `FlutterWindowMetricsEvent` gained size-constraint members, which this port does not negotiate, so the three initializers in `engine.cc` now declare their absence explicitly.

## 3. Bounds-check the flags struct — the interesting one

**`FlutterSemanticsFlags` is versioned independently of the node that points at it.** It carries its own `struct_size` and grows on its own schedule. Commit 2 gated the node struct and — precisely because that felt like the whole job — missed that the flags struct hanging off it had *also* grown a member. `is_accessibility_focus_blocked` is absent at 3.38.3, so reading it unconditionally was an out-of-bounds read on the flags allocation against **every deployed engine**.

That was found by auditing the version floor, not by any test. It is now gated with a `static_assert` on the offset arithmetic.

### Why the tests are shaped the way they are

Both regression tests size the allocation to the **actual deployed struct** rather than a synthetic short one, and place it at the *end* of a heap allocation — so an unguarded read faults instead of picking up adjacent live memory that happens to look plausible. Verified by deliberately removing each gate:

```
ERROR: AddressSanitizer: heap-buffer-overflow
SUMMARY: heap-buffer-overflow accessibility_tree.cc:75 in AccessibilityNode::Update()
SUMMARY: heap-buffer-overflow semantics_translator.cc:170 in FromStruct
```

**The generalizable rule: a gate on a containing struct says nothing about a struct it points to.** Each embedder struct the port consumes needs its own audit. `FlutterSemanticsCustomAction2` and `FlutterSemanticsUpdate2` are next if they ever grow.

## On the dispatch API

`FlutterEngineSendSemanticsAction` is declared by the rolled header **and exported by 3.38.3, with `view_id`** — so it works at the floor. Nothing here adds it to the dlsym table regardless: `libflutter_engine.cc` resolves with `require()`, which aborts load on a missing symbol, so a hard dependency would turn any engine below the floor from degraded into non-starting. A caller wanting it should resolve it optionally and fall back to the `DispatchSemanticsAction` entry already present.

`GetRole()` still reports the narrow numeric encoding the AccessKit bridge consumes, so that path is behaviorally unchanged — worth a look from anyone who builds with `ENABLE_ACCESSKIT`, since `find_package(ACCESSKIT)` fails on my host and that block never compiled locally.

## Testing

- 23 new cases in `test_case_accessibility_tree.cc` (7 → 30): action bitmask, transform, identifier retention and replacement, custom-action resolution and string ownership, null-pointer and null-string handling, the no-prune property, `flags2`-over-enum precedence, tristate semantics, null-`flags2` fallback, post-ceiling actions, both ASan bounds tests, and dump output.
- 22/22 ctest green; accessibility and translator suites clean under ASan + UBSan.
- clang-tidy clean on changed files; clang-format applied.
- `BUILD_ACCESSIBILITY=ON` builds warning-free.

**Suggested follow-up:** a CI leg pinned to a 3.38.3 engine, so the degradation paths are executed rather than reasoned about. Every conclusion above about 3.38.3 comes from reading its header, not from running against it.

Unrelated, not addressed here: building a single unit-test target fails because `homescreen_ut_shell` does not depend on the wayland-cxx-scanner generation targets. Generating `shell/wayland-protocols/*.hpp` first works around it.

### DIFF
--- a/shared/include/ihs/ihs_semantics.h
+++ b/shared/include/ihs/ihs_semantics.h
@@ -1,0 +1,428 @@
+/*
+ * Copyright 2026 Toyota Connected North America
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * ihs_shared semantics hub: the single fan-out point for the Flutter semantics
+ * tree, and the single funnel for actions dispatched back into it.
+ *
+ * The shell mirrors each FlutterSemanticsUpdate2 into a mutable tree on the
+ * platform thread. That tree is never exposed. Instead, at every batch
+ * boundary the hub publishes an immutable, refcounted IhsSemanticsSnapshot;
+ * consumers read snapshots and never touch the mutable tree, so a slow
+ * consumer cannot stall the platform thread.
+ *
+ * This surface is deliberately free of any dependency on embedder.h. A
+ * consumer may be an out-of-tree FFI plugin that has no Flutter headers at
+ * all, so roles, states and actions are re-expressed here rather than
+ * forwarded. See docs/PLUGIN_ABI.md for the boundary rules.
+ *
+ * Lifecycle of one consumer:
+ *   1. Fill an IhsSemanticsConsumerDesc (name, allowed actions, notify fd) and
+ *      call ihs_semantics_register. Registration is process-global and may
+ *      happen at any time -- a client attaching mid-session needs no restart.
+ *   2. The hub reference-counts registrations to drive semantics on and off in
+ *      the engine. With no consumer registered the engine is never asked to
+ *      produce semantics at all, so a build with adapters compiled in but none
+ *      configured pays nothing.
+ *   3. Wait on the notify fd. Each readable event means "a newer generation
+ *      exists"; events are edge-coalesced, so a consumer that wakes once for
+ *      three updates has missed nothing. Always fetch the latest snapshot on
+ *      wake rather than trying to track individual updates.
+ *   4. ihs_semantics_acquire_snapshot for the current tree; release it when
+ *      done. Acquire/release are lock-free and callable from any thread.
+ *   5. ihs_semantics_dispatch to invoke an action. The hub serializes onto the
+ *      platform thread, enforces the consumer's action_allow_mask, and traces
+ *      the dispatch with the consumer name so "which actor did this" is
+ *      answerable after the fact.
+ *   6. ihs_semantics_unregister drains any in-flight dispatch for that
+ *      consumer before returning.
+ *
+ * Notification is by file descriptor, not callback, on purpose: one consumer
+ * that blocks in a callback would delay notification of every other, and
+ * "must not block" is a contract that erodes. Consumers poll in their own
+ * event loop instead.
+ *
+ * Threading: register/unregister and dispatch are callable from any thread.
+ * Snapshot acquire/release are callable from any thread and are lock-free. The
+ * done callback passed to dispatch is invoked on the platform thread. Snapshot
+ * contents are immutable for the lifetime of the reference held.
+ */
+
+#ifndef IHS_SEMANTICS_H_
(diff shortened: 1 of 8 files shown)


jwinarske:
Added a fourth commit: `shared/include/ihs/ihs_semantics.h`, the consumer-facing ABI for the semantics hub.

**Header only — no implementation, no exported symbols, so the `abidiff` gate is untouched.** Posting the interface first so the shape can be argued with before anything is built against it.

Three decisions worth disagreeing with if you disagree:

**No `embedder.h` dependency.** A consumer may be an out-of-tree FFI plugin with no Flutter headers at all, so roles, states and actions are re-expressed rather than forwarded — including the action bit *values*, which are this ABI's own. Costs a translation layer; buys a surface that does not move when the engine's own flag representation does. Given the last few commits on this branch were about exactly that kind of movement, it seemed worth the cost.

**States are tristates, not bools.** A checkbox that is unchecked and a node where "checked" is meaningless are different, and a bool cannot say which one it is holding. This mirrors what the newer engine flags struct now expresses and the older enum could not.

**Notification is an fd, not a callback.** One consumer blocking in a callback delays every other consumer's notification, and "must not block" is a contract that erodes. Events are edge-coalesced, so waking once for several updates loses nothing as long as consumers always fetch the latest snapshot rather than tracking individual updates.

The per-consumer `action_allow_mask` is the arbitration point when a programmatic driver and a screen reader share one tree. `IHS_SEMANTICS_ACTION_NO_A11Y_FOCUS` withholds the two accessibility-focus actions — those move a screen reader's cursor, and something driving the UI on its own initiative should not be able to yank it away from someone reading the screen.

Two smaller things:

- `identifier` is documented as **empty rather than absent** on engines predating the field. A consumer otherwise cannot distinguish "the app did not annotate this" from "this engine cannot report it" and would draw the wrong conclusion. Relevant today: at the 3.38.3 floor it is empty on every node.
- `IHS_SEMANTICS_ACTION_ALL` is spelled as the OR of the named bits rather than a hex literal, so it cannot drift when an action is added. I had written it as a literal first, verified it by hand, and then decided a constant that needs verifying is a constant that will eventually be wrong.

Compiles clean as C99, C11 and C++17 under gcc and clang with `-pedantic -Wall -Wextra -Wconversion`.

PR #415, opened 2026-08-07
shared: implement the semantics hub

jwinarske:
**Stacked on #414** — base is `jw/a11y-node-enrichment`, so this diff shows only the hub. Merge #414 first; GitHub will retarget this to `v3.0` automatically.

Implements the `ihs_semantics.h` ABI added in #414: snapshot publication, the consumer registry, fd notification, and the action dispatch funnel. Adds the shell seam as `ihs_semantics_host.h`, following the existing `platform_view.h` / `platform_view_host.h` split — a consumer sees only the plugin header.

No shell wiring yet and no AccessKit migration; see "What is not here".

## Design points worth review

**Publication is whole-tree, not incremental.** A snapshot is immutable once published, so there is nothing to patch, and consumers never reassemble a consistent view out of deltas. The shell hands over nodes addressed by id; the hub resolves child ids to indices once at publish, so walking costs no lookups. A child id with no node in the batch is **dropped** rather than published as a dangling reference — every published index is dereferenceable.

**A mutex, not a lock-free protocol.** It is held only to swap a pointer or touch the registry — never across consumer work or a call into the shell. The property the design actually needs is "no consumer can stall the platform thread, and none can stall another", and that holds. Lock-freedom for its own sake would buy nothing here and cost a reclamation scheme. Snapshot building and fd writes both happen *outside* the lock, since neither should be able to hold the registry closed.

**Mask is checked before node lookup.** Checking it after would let a caller probe the tree with actions it is not permitted to invoke, reading the answer out of the difference between `ERR_DENIED` and `ERR_NO_SUCH_NODE`. There is a test for the ordering specifically.

**Actions the node does not offer are refused.** The framework silently drops an action the widget registered no handler for, so passing it through would make the call vanish. Refusing means the caller learns something.

**Engine semantics follows the consumer count**, on the 0→1 and 1→0 transitions only — a process with no consumer never asks the engine to produce semantics at all.

## Two header adjustments, flagged rather than buried

I changed two threading claims in `ihs_semantics.h` to match what is implemented, rather than the reverse:

- Acquire was documented "lock-free"; it is now documented as O(1) contending only with publish. The honest statement, and the guarantee consumers care about, is that no consumer blocks another.
- The done callback was documented as firing on the platform thread; it now fires on the caller's thread once a dispatch is *accepted*. This also made it natural to spell out that acceptance is not execution and certainly not a visible UI change — a caller wanting to know what changed must compare generations.

Neither changes the surface. Say the word if you'd rather have the async contract and I'll queue dispatches instead.

## Testing

22 cases driving the exported C ABI with a mock host — same approach as `ihs_pv-test`, so no engine, backend, or app bundle is involved and it runs anywhere. Covers publish validation, id→index resolution, dangling child ids, null-string normalization, snapshot lifetime across republication, generation ordering, custom action resolution, refcounted enable/disable, mask denial ordering, unsupported/unknown actions, payload forwarding, dispatch after unregister, missing host, and fd notification.

A concurrent test churns 200 publishes against four reader threads that dereference snapshot contents — the refcount protocol is the whole reason a consumer can read off the platform thread, so it is exercised rather than trusted by inspection.

- 23/23 ctest green.
- Clean under **ASan, UBSan and TSan** (no data races reported).
- clang-tidy clean; clang-format applied.
- 14 `ihs_semantics_*` symbols export through the version script, with no C++ symbol leakage.
- `ihs_semantics_host.h` compiles as strict C99.

**ABI gate:** this adds exported symbols only. `abidiff --no-added-syms` fails on bit 3 (incompatible) only, and the workflow comments say additive changes pass, so the committed baseline should not need regenerating. Worth a look from someone who has watched that gate behave.

## What is not here

- **Shell wiring** — nothing calls `ihs_semantics_publish` or installs a host yet. `AccessibilityTree` has the enriched nodes from #414; connecting them is the next commit.
- **AccessKit migration onto the hub.** `find_package(ACCESSKIT)` fails on my host so that block never compiles locally, and I would rather not refactor code I cannot build.


jwinarske:
Added the shell wiring (`e2f41b93`), so the hub is now fed and served end to end rather than sitting inert.

- The engine installs itself as the hub's host at bring-up.
- `onSemanticsUpdateCallback` publishes each batch at the boundary — the only point the mirror is known consistent.
- Dispatches come back through the platform task runner via `asio::post` on the runner's strand.
- **`UpdateSemanticsEnabled(engine, true)` at bring-up is gone.** The hub reference-counts consumers and asks through the host, so a `BUILD_ACCESSIBILITY=ON` build with nobody listening now costs the engine nothing. That was the point of the refcount, but it does change existing behavior — worth a look.

## Three things worth review

**`SendSemanticsAction` is resolved optionally, not required.** `libflutter_engine.cc` already had an `optional` resolver alongside `require`, so this uses it. Requiring the symbol would turn an engine older than the vendored header from *degraded* into *won't start* — the §2.1.1 concern, now load-bearing rather than theoretical. When absent, the deprecated `DispatchSemanticsAction` is used; it has no `view_id`, so a dispatch aimed at any view but the implicit one is **dropped with a warning** rather than landing somewhere it wasn't aimed.

**All four rect corners are mapped, not two.** Under rotation or skew the transformed min/max corners aren't the transforms of the original min/max, so a two-corner bound is quietly wrong for exactly the cases that need it. There's a 90°-rotation test pinning it.

**The translator gained three-valued state.** Publishing the collapsed bools would have discarded the distinction the newer flags struct exists to draw — a node where enablement is meaningless and one that's disabled are both `false` as a bool. The bools remain for callers that only ask "is it on", documented as lossy. I also fixed `selected`/`focused`/`required` to keep their tristate through the flag view, since `flags2` reports those as tristates too and my first pass flattened them.

## Testing

14 new cases (24 ctest total) over the composition arithmetic, the root-first walk, tristate round-tripping, and both directions of the action mapping. Including one that asserts **every** action the hub defines has a framework equivalent — so a bit added to one side without the other fails the build's tests rather than silently becoming undispatchable.

24/24 ctest, clean under ASan/UBSan, clang-tidy clean, `BUILD_ACCESSIBILITY=ON` warning-free.

## Still not done

**AccessKit migration onto the hub.** Unchanged reasoning: `find_package(ACCESSKIT)` fails here so that block never compiles locally. It still reads the mirror directly rather than the snapshot. Happy to do it if someone with an accesskit build can verify, otherwise I'd rather leave it than refactor blind.

Also worth noting: nothing registers as a consumer yet, so in practice semantics stays *off* until something does. That's correct per the refcount design, but it means this branch alone produces no observable behavior change in a running shell — the first consumer is what lights it up.

PR #426, opened 2026-08-10
test: add the OSGi multi-bundle vkms harness, and stop hardcoding the DRM card

jwinarske:
Two test-infrastructure changes. The second is a **safety fix to existing
code** and is worth reading first.

---

## 1. `test: stop hardcoding the DRM card in the KMS harnesses`

`multi_view_single_engine.sh` defaulted `DRM_DEVICE` to `/dev/dri/card1`. That
is not a portability nit: these harnesses **take DRM master and drive a
modeset**, so on any machine where `card1` is the GPU rather than vkms — the
common case on a developer desktop — running the harness without arguments goes
after the real display.

Card numbering is not stable to begin with. vkms lands wherever it lands
relative to any real GPU and moves as other DRM drivers load, so no numeric
default is correct anywhere except the machine it was written on. The connector
defaults had the same problem in the *other* direction: `HDMI-A-1`/`DP-1` only
exist on a real GPU while vkms advertises `Virtual-N`, so the default node and
the default connectors could never both be right.

There is now no numeric default anywhere, and the safety property is that
**auto-detection resolves vkms and nothing else**. Targeting real hardware
requires naming the node explicitly, which is the operator saying so out loud.
With no vkms present the harness refuses rather than guessing. Connector names
are read off whichever card resolved, with Writeback filtered out — a capture
target, not a scanout one.

Detection moves to `test/lib/drm_card.sh` rather than being copied a third
time. It is subtle enough to be worth having once: vkms is identified by its
connector signature (`cardN-Virtual-M`) rather than by the driver symlink,
whose name has already changed once across kernel versions (`platform` →
`faux_driver`) and would silently stop matching on the next rename. Three
copies would mean three places to miss. `drm_kms_vkms.sh`, where the method
originated, now sources it too.

Verified on a box whose `card1` is a real GPU with vkms not loaded — the exact
configuration the old default would have driven a modeset on:

| | |
|---|---|
| `drm_kms_vkms.sh --check` | unchanged (stops at the module check) |
| `multi_view_single_engine.sh` | refuses; previously would have used `card1` |
| `osgi_multi_bundle.sh` | skips honestly, exit 0 |
| explicit `DRM_DEVICE=…` | still works, reports the real connectors |
| `ihs_connectors_of card1` | `DP-1 HDMI-A-1`, `Writeback-1` excluded |

`drm_kms_vkms.sh --check` got particular attention since CI runs it as a
pre-flight.

---

## 2. `test: add the OSGi multi-bundle vkms harness`

The integration tier of the `BundleEngineManager` split. Unit tests can only
assert the framework's central claim — **two bundles, two engines, critical
first** — against a fake host. This proves it on real KMS.

It is the exact complement of `multi_view_single_engine.sh`. That harness proves
ONE engine can drive TWO outputs and **fails** if a second engine appears; this
one **requires** the second engine, because an OSGi bundle *is* a separate
engine and every isolation claim the framework makes rests on that. The two key
off the same log line from opposite directions.

Five assertions, each covering a way the framework could look correct while
being wrong:

| | |
|---|---|
| **B1** two engines | A shell that collapsed both bundles onto one engine would still paint. |
| **B2** both outputs | Per-bundle `[osgi.bundles.backend]` honored, not the first bundle's winning for all. |
| **B3** critical first | Bundles are declared **normal-first** in the generated config, so a pass proves the orchestrator *reordered* them rather than following file order. |
| **B4** clean log | No error/critical lines. |
| **B5** page flips | Both CRTCs actually presenting — construction without presentation satisfies B1 and B2 while showing nothing. |

The generated config has **no `[[view]]`**, only `[[osgi.bundles]]` — the shape
of a pure-OSGi deployment, where every surface belongs to a bundle. That is a
requirement the harness places on the implementation rather than an incidental
detail: the bundles must themselves become the views.

### Wired into CI now, inert until it matters

It skips honestly rather than failing when the binary lacks `ENABLE_OSGI` (the
default), when no vkms card is present, or when the card has fewer than two
scanout connectors. That is what lets it be added to
`vkms_harness_integration.sh` today, where it reports a skip in the job summary
and goes live by flipping `ENABLE_OSGI` in the vkms-smoke build step once the
host implementation lands. Wiring it up front means it cannot be forgotten at
the point it starts to matter.

### What is and is not verified

Verified: syntax; the missing-env guards; the non-DRM-binary guard; the
non-OSGi skip; the no-vkms skip; `--check` against a real DRM+OSGi binary;
explicit `DRM_DEVICE` opt-in; connector auto-detection.

**Not verified: the five assertions themselves.** They cannot run until the
host implementation exists, and a live-fire run needs vkms, which needs root.
This harness is the executable specification for that slice.

### DIFF
--- a/scripts/vkms_harness_integration.sh
+++ b/scripts/vkms_harness_integration.sh
@@ -82,5 +82,21 @@ run_harness "present census" \
 run_harness "event-driven input harness" \
     env IDLE_CTXT_MAX=150 "${IVI_SRC}/test/input_event_driven.sh"
 
+# ---------------------------------------------------------------------------
+# OSGi multi-bundle: two bundles -> two engines -> two connectors, critical
+# first. Needs two connectors on one card and a binary built with
+# ENABLE_OSGI=ON; the harness reports an honest skip when either is missing
+# rather than failing, so it is wired in here from the start and goes live by
+# flipping ENABLE_OSGI in the build step above it.
+# ---------------------------------------------------------------------------
+# DRM_DEVICE and the connector names are deliberately NOT passed: card
+# numbering is not stable (vkms lands wherever it lands relative to any real
+# GPU) and connector names are card-specific (Virtual-N on vkms, HDMI-A-1/DP-1
+# on a real GPU). The harness auto-detects vkms by its connector signature and
+# reads the connectors off that card, so nothing here has to guess. It skips
+# honestly when there is no vkms card or the binary lacks ENABLE_OSGI.
+run_harness "osgi multi-bundle" \
+    env SOFTWARE_RENDER=1 "${IVI_SRC}/test/osgi_multi_bundle.sh"
+
 echo "vkms harness integration: overall rc=${rc}"
 exit "${rc}"
--- a/test/drm_kms_vkms.sh
+++ b/test/drm_kms_vkms.sh
@@ -84,23 +84,9 @@ log() {
     echo "==> $*"
 }
 
-find_vkms_card() {
-    # Identify the vkms card by its connector signature. vkms always
-    # advertises connectors named "cardN-Virtual-M"; real GPUs don't.
-    # This is more reliable than matching the driver symlink, whose name
-    # has changed over kernel versions (platform → faux_driver).
-    local c card
-    for c in /sys/class/drm/card[0-9]*; do
-        # Skip the connector/encoder child nodes themselves.
-        [[ -d "$c/device" ]] || continue
-        card="$(basename "$c")"
-        if compgen -G "/sys/class/drm/${card}-Virtual-*" >/dev/null; then
-            echo "/dev/dri/${card}"
-            return 0
-        fi
-    done
-    return 1
-}
+# Card/connector discovery is shared with the other KMS harnesses.
+# shellcheck source=test/lib/drm_card.sh
+source "$(dirname "${BASH_SOURCE[0]}")/lib/drm_card.sh"
 
 ensure_vkms_loaded() {
     if lsmod | awk '{print $1}' | grep -qx vkms; then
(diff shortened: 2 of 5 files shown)


jwinarske:
Updated with two commits from the first hardware run, on an rpi4 (trixie, vc4, `DSI-1` + `HDMI-A-1`). The harness now passes **5/0** there:

```
liveness          pass — still running after 4s
B1-two-engines    pass — 2 distinct engine indices
B2-both-outputs   pass — DSI-1 and HDMI-A-1 both up
B3-critical-first skip — needs ACTIVATOR=1
B4-clean-log      pass — no error/critical lines
B5-page-flips     pass — 864 flip/atomic ioctls observed
```

Three fixes, each one a thing only a real rig produced:

**Bundles cannot ship their own engine library.** `LibFlutterEngine::Load()` is one-shot and keyed on the path it first binds, so a second bundle asking for *any* different path — including the bare system name — is refused and its `FlutterView` aborts. Two bundles each carrying `lib/libflutter_engine.so` cannot coexist **however identical the files are**, which is exactly what copying one bundle twice produces. The harness hoists a single shared copy. This is a deployment constraint that follows from the one-VM-per-process premise but had never been written down.

**B1 was keying on a log line that is never emitted.** `Engine running` occurs **zero** times in a run that demonstrably brought up two engines on two connectors, so the assertion reported a false failure. It now counts distinct engine indices from `(N) Loading AOT` and `view N is on output`.

> Worth flagging separately: `multi_view_single_engine.sh` keys its own "no second engine" check on that same absent string, so **that assertion currently cannot fail**. I have not touched it — it is someone else's assertion and fixing it changes what that harness reports. Happy to do it in a follow-up if you want.

**Connector discovery.** Only `connected` connectors are listed now — a disconnected one still appears in sysfs, so pinning a bundle to it produced a modeset failure deep in the backend instead of an obvious "there is no second display here". And `ihs_find_vkms_card` takes a minimum connector count, because `modprobe vkms` creates a single-output instance and `vkms_dual.sh` then adds a two-output one; a bare first-match search found the wrong card.

**`ACTIVATOR` gate.** The critical-first guarantee is "critical is ACTIVE before normal starts", and ACTIVE arrives from the bundle's Dart activator over `dev.osgi/bridge` — `grep -c dev.osgi/bridge` is **0** on a real run, because an ordinary Flutter app never sends it. The critical wait therefore times out and tears the bundle down, failing B1/B2 for reasons unrelated to what they test. With `ACTIVATOR=0` (default) both bundles run at normal priority and B3 skips saying exactly that; set it to `1` once an activator-capable bundle exists.

The failure path *is* proven on hardware even though the success path is not: on timeout the bundle went `STARTING → STOPPING → RESOLVED`, its view was torn down, and startup continued to the next bundle.

PR #439, opened 2026-08-12
shared: encode StandardMessageCodec action arguments

jwinarske:
Semantics actions that carry an argument — `SetText`, `ScrollToOffset`, `SetSelection` — need it as a StandardMessageCodec message, and nothing in `ihs_shared` could produce one. This adds the three shapes those actions use: a `String`, a `Float64List`, and a `Map<String, int>`.

This unblocks two separate things that are stuck on the same missing piece: the parameterized `ui_*` tools, and AccessKit's `SetValue` write path (an assistive technology can currently read a scroll position but not set one, because `SetValue` maps to `ScrollToOffset` and there was no payload to send).

## Why not the vendored codec

Flutter vendors a complete implementation in `third_party`, and this is deliberately not it.

`ihs_shared` links **only** `Threads::Threads` — no Flutter library. It uses Flutter's public *C* headers for the desktop-bridge forwarders and nothing else. The vendored codec is C++ over `flutter::EncodableValue`, so using it here means linking `libflutter.a` into the library whose stated premise is *"No C++ types, exceptions, or RTTI cross the boundary"* (`docs/PLUGIN_ABI.md`). That is a large trade for three values.

The alternative — encoding shell-side, where `libflutter` is already linked — would mean the hub ABI carrying typed Flutter data instead of opaque bytes, putting Flutter's data model into a surface that is deliberately free of it.

So: a small encoder here, internal to `ihs_shared`, not installed and not exported through the `ihs_*` ABI. The tests compile the source directly rather than linking it.

## What makes a second implementation safe to keep

A reimplementation of someone else's format can drift from it silently — that is R-9 in the plan, "codec mis-encoding drops action args silently framework-side". The mitigation is the point of this PR: **every encoding is decoded again with Flutter's own vendored serializer and compared.** The test driver can link `libflutter` even though `ihs_shared` cannot, so a divergence fails here rather than reaching the framework as a dropped argument.

10 tests, covering the parts actually worth doubting rather than the happy path:

- all three **length-prefix forms** (`<254`, the 16-bit sentinel, the 32-bit one) — a reader that disagreed about which was used would misread everything after it
- **multi-byte UTF-8**, where the length is bytes and not characters
- **alignment padding** for a typed list, which is measured from the start of the message rather than from the list; the assertion pins the exact byte count, so getting the origin wrong fails
- zero and negative values carried rather than dropped — a collapsed selection and a zero scroll offset are ordinary values

## One thing found on the way

Flutter's own C++ codec is **asymmetric for an empty typed list**: `WriteVector` returns before writing alignment when the count is zero, but `ReadVector` calls `ReadAlignment` unconditionally. So its reader walks past the end of what its writer just produced, and logs `Invalid read in StandardCodecByteStreamReader`.

This encoder matches the **writer**, because the writer is what the wire format is — the Dart side decodes these bytes, not Flutter's C++ reader. That case is asserted as exact bytes rather than by round trip, with a comment recording why, so nobody later "fixes" the encoder by adding padding to silence a warning from a reader nothing in this path uses. `ScrollToOffset` always carries two elements, so it is academic here regardless.

## Scope

**Not yet wired to a caller.** The tools that carry these arguments (`ui_set_text`, `ui_scroll_to`) and the AccessKit `SetValue` mapping come next; this is the piece both need, landed on its own so the format can be reviewed against Flutter's without a behaviour change alongside it.

28/28 tests, clang-tidy clean, `scripts/format.sh --check` and `gen_config_reference.py --check` both pass.

### DIFF
--- a/shared/src/standard_message_encode.cc
+++ b/shared/src/standard_message_encode.cc
@@ -1,0 +1,112 @@
+/*
+ * Copyright 2026 Toyota Connected North America
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "standard_message_encode.h"
+
+#include <cstring>
+
+namespace ihs::codec {
+namespace {
+
+// Wire tags, in the order Flutter's serializer defines them. Written out
+// rather than referenced because the enum lives in a header this library does
+// not include; the round-trip tests are what keep the two agreeing.
+constexpr uint8_t kString = 7;
+constexpr uint8_t kFloat64List = 11;
+constexpr uint8_t kMap = 13;
+constexpr uint8_t kInt32 = 3;
+
+// Sizes below 254 are a single byte; the two sentinels buy 16- and 32-bit
+// lengths. A reader that got this wrong would misread everything after it, so
+// it matches the serializer exactly rather than always using the wide form.
+void WriteSize(std::vector<uint8_t>& out, const size_t size) {
+  if (size < 254) {
+    out.push_back(static_cast<uint8_t>(size));
+  } else if (size <= 0xffff) {
+    out.push_back(254);
+    const auto value = static_cast<uint16_t>(size);
+    const auto* bytes = reinterpret_cast<const uint8_t*>(&value);
+    out.insert(out.end(), bytes, bytes + sizeof(value));
+  } else {
+    out.push_back(255);
+    const auto value = static_cast<uint32_t>(size);
+    const auto* bytes = reinterpret_cast<const uint8_t*>(&value);
+    out.insert(out.end(), bytes, bytes + sizeof(value));
+  }
+}
+
+// Typed lists are padded so their elements start on a natural boundary,
+// measured from the beginning of the message rather than from the list. The
+// decoder skips the same number of bytes by the same rule, so getting the
+// origin wrong shifts every element.
+void WriteAlignment(std::vector<uint8_t>& out, const size_t alignment) {
+  const size_t mod = out.size() % alignment;
+  if (mod == 0) {
+    return;
+  }
+  out.insert(out.end(), alignment - mod, 0);
+}
+
+void WriteInt32(std::vector<uint8_t>& out, const int32_t value) {
+  out.push_back(kInt32);
+  const auto* bytes = reinterpret_cast<const uint8_t*>(&value);
+  out.insert(out.end(), bytes, bytes + sizeof(value));
+}
+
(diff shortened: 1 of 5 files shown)


jwinarske:
Closing this: it should not have been written as a new encoder.

**plan.md §184 already settled the approach** — "Encode with the vendored `flutter::StandardMessageCodec` (`third_party/flutter/shell/platform/common/client_wrapper/standard_codec.cc`) — no new serialization code." This PR is new serialization code, and I departed from that without flagging it.

The justification given here — that using the vendored codec would violate the flat C ABI premise — does not hold. `shared/ihs_shared.map` exports only `ihs_*` and `FlutterDesktop*` with `local: *`, so C++ symbols from `libflutter.a` stay local and cross no plugin boundary. The premise in `docs/PLUGIN_ABI.md` is about the exported surface, not about internals.

## What to do instead, if this work resumes

**Encode shell-side.** The shell already links `libflutter` and has the vendored codec available, and its dispatch callback already knows which action it is handling, so it can build the payload there. The hub keeps passing opaque bytes, `ihs_shared` stays free of any Flutter library, and no CMake reordering is needed.

The alternative — linking the codec into `ihs_shared` — additionally needs `add_subdirectory(third_party)` (root `CMakeLists.txt:114`) moved above `add_subdirectory(shared)` (line 109), since the `flutter` target does not exist yet when `shared` is configured. The comments at lines 71–78 record existing ordering constraints, so that is not a free change.

## What is worth keeping from this branch

The round-trip tests. They decode with Flutter's own vendored serializer and cover the parts worth doubting: all three length-prefix forms, multi-byte UTF-8 where the length is bytes rather than characters, and typed-list alignment measured from the start of the message. They are just as useful against the vendored codec, where they satisfy R-9's "codec round-trip tests as P3 exit criterion" without guarding a reimplementation.

Also recorded here so it is not rediscovered: Flutter's own C++ codec is **asymmetric for an empty typed list** — `WriteVector` returns before writing alignment when the count is zero, while `ReadVector` consumes alignment unconditionally, so its reader walks past the end of what its writer produced and logs `Invalid read in StandardCodecByteStreamReader`. The writer is the wire format, so anything encoding for Flutter should match the writer and ignore that warning.

## On performance

Raised and checked: this encoder cannot replace any existing code. `shell/` has no explicit StandardMessageCodec encode sites — encoding lives inside the client_wrapper channel classes, which handle arbitrary `EncodableValue` trees — and the hot-ish path (`platform_view_touch.cc`) is a *decoder*, which this has no side for. It handles three fixed shapes; generalizing it would be reimplementing the vendored codec wholesale.

The genuine performance opportunity is elsewhere and is not this format: pushing the semantics **tree** to Dart (large, structured, potentially animation-rate — plan R-2), and the T3 manifest channel (plan §202, "the only new wire format in this plan"). Both are wires ivi-homescreen owns on both ends, so both belong on `native_comms` Channel E (Glaze/BEVE-Lite over `kExternalTypedData`), where field elision is ~3x on struct-shaped data and the transport is genuinely zero-copy. Measured on this PR's own payloads for comparison: BEVE-Lite is 3x smaller on `SetSelection`, a tie on `ScrollToOffset`, and *larger* on a short `SetText` string, because its fixed uint64 length prefix costs more than a tag plus a one-byte size.

None of that applies to semantics action arguments themselves, which are decoded framework-side by Flutter and so are not ours to format at any speed.

Branch `jw/standard-message-encode` is left in place (CI green) should any of the above be wanted.

PR #447, opened 2026-08-12
shared: stream notifications to clients over SSE

jwinarske:
The notification path (#444) had nowhere to deliver. This is the server-to-client half of the transport: a `GET` asking for `text/event-stream` opens a stream, and provider changes arrive on it as `notifications/resources/updated` and `notifications/tools/list_changed`.

With this, a client no longer has to poll `ui_snapshot` to notice the UI changed. That completes P3's transport.

## `initialize` now advertises what it can deliver

`listChanged` and `subscribe` were deliberately withheld while nothing could send them — a client told to expect notifications that never arrive *waits* instead of polling, which is worse than being told nothing. They are advertised now because they are real.

## A stream is not a request

The transport served every connection inline on the accept thread, which is correct for a request/response exchange and wrong for a stream: one connected client would occupy the loop and stall everyone else for as long as it stayed. So the accept loop now hands the descriptor to a stream registry and stops closing it, and requests continue to be served alongside — asserted by a test that issues a `ping` while a stream is open.

**Streams are non-blocking.** Delivery happens on the registry's watcher thread, so a client that stopped reading would otherwise stall every *other* provider's notifications too. Such a client is dropped rather than waited for: a partial frame desynchronises a stream and there is no way to resend one.

**Ordering at teardown.** The sink is installed after the accept loop is up, and removed before the streams are closed. Uninstalling joins the registry's watcher, so no delivery can be in flight into a descriptor that is being closed.

## Testing

**Driven with a real client over the socket**, publishing on a timer:

```
events received: 10
methods: ['notifications/resources/updated', 'notifications/tools/list_changed']
  {"jsonrpc":"2.0","method":"notifications/tools/list_changed"}
  {"jsonrpc":"2.0","method":"notifications/resources/updated","params":{"uri":"ui://"}}
```

6 new unit tests (27 in the suite): the handshake returns `text/event-stream` with no `Content-Length`, a plain `GET` is `406` rather than handed a stream it will not read, requests are still served while a stream is open, several streams may be open at once, and stopping closes them — a client sees end-of-stream, which is how it learns the server is gone, since there is no goodbye event.

**Two existing tests asserted the old contract and are updated rather than removed.** One pinned that `initialize` withholds `listChanged`; the other that `GET` is refused. Both were right before and wrong now, and they caught the change — the second is rewritten to assert that genuinely unsupported verbs (`PUT`, `DELETE`, `HEAD`) still get `405`, so "wrong verb" stays distinguishable from "right verb, wrong media type".

28/28 across the suite, clang-tidy clean, both repo gate scripts pass.

## Not included

Resource *subscriptions* are advertised and accepted, but every stream currently receives every notification rather than only the URIs it subscribed to. The registry already tracks subscriptions (`ihs_mcp_host_subscribe`); filtering per stream is a follow-up, and over-delivery is the safe direction — a client can ignore an update it did not ask for, but cannot recover one it never received.

### DIFF
--- a/test/integration/mcp_drive_test/README.md
+++ b/test/integration/mcp_drive_test/README.md
@@ -1,0 +1,93 @@
+# mcp_drive_test
+
+Integration test app for **driving ivi-homescreen over MCP**, plus a driver
+that acts through the socket and checks what the application actually did.
+
+Every other MCP test in this repo runs against a mock host. Those prove a
+request reaches the shell; they cannot prove Flutter did anything with it,
+because there is no Flutter on the other side. This fixture is that other
+half:
+
+```
+MCP client (tools/drive.py)
+  -> Unix socket -> HTTP -> JSON-RPC        (transport)
+  -> ihs_mcp_host -> semantics provider     (routing)
+  -> ihs_semantics_dispatch                 (funnel: mask, attribution)
+  -> FlutterEngineSendSemanticsAction       (embedder)
+  -> SemanticsOwner.performAction           (framework)
+  -> the closure the widget registered      <- only this fixture reaches here
+```
+
+## The app reports through MCP
+
+Every handler writes into a status line that is itself a semantics node, so
+the driver reads results back over the same surface it used to act. The read
+path is therefore the assertion channel for the write path, and neither can
+pass on its own.
+
+The status value is semicolon-separated `key=value`, because one field is
+text the driver chose and the realistic `ui_set_text` case contains a space:
+
+```
+last=tap:save; events=3; text=Sarah Connor; opaque=16.0,412.0,240.0,80.0
+```
+
+`opaque` is where the unannotated region landed after layout. The app reports
+its own geometry so the driver can aim `ui_tap_at` without hard-coding a
+layout that breaks the first time a font or a screen size changes.
+
+## Checks
+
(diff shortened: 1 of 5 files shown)


jwinarske:
Closed automatically, not abandoned — and the code **is** merged.

I branched `jw/mcp-inspector` off `jw/mcp-sse` instead of off `v3.0`, so #448 carried this PR's commit. Merging #448 landed SSE on `v3.0`, which left this branch with nothing to merge and GitHub closed it.

```
b75234ec  Merge pull request #448 from …/jw/mcp-inspector
  ├─ 64560aab  shared: stream notifications to clients over SSE     <- this PR
  └─ b6cee5d4  scripts: add a live viewer for the semantics tree
```

Nothing is lost: the commit is on `v3.0` with its message and authorship intact, and the transport serves `text/event-stream`, advertises `listChanged` and `subscribe`, and pushes both notification kinds.

What is lost is provenance — the reasoning for the concurrency change (a stream cannot be served inline on the accept thread), for non-blocking streams (delivery runs on the registry watcher, so a stalled client must not block every other provider's notifications), and for updating the two tests that pinned the old contract, all lives here on a PR that never merged. Recording that so the reasoning is findable from the code.

My mistake in branching, not a merge accident.

PR #484, opened 2026-08-17
fix: publish a tree per Flutter view rather than merging them

jwinarske:
Second of three. #483 has merged, so this is rebased onto `v3.0` and stands alone.

## The defect

One engine can drive several Flutter views: the DRM backend attaches an additional view per extra output on the same card (`flutter_view.cc`, `Engine::AddView`). Their semantics were merged into one mirror, and **the merge destroyed them**.

Every view's root is node 0 — the framework defines it that way (`SemanticsNode.root()`). Two views applied to one tree write over each other at the root. Replacing the root's children detaches the previous view's subtree, and a detached node is **pruned rather than kept**. The publisher then walks from node 0 and emits what it can reach.

So the losing view's nodes do not come out stale or misparented. They come out **absent**, with nothing anywhere reporting that a view stopped being described.

The batch already said which view it belonged to. `onSemanticsUpdateCallback` discarded it.

## The fix

Each view gets its own tree and its own hub source, named for the view it belongs to, registered the first time that view publishes.

Actions come back through the source they were read from: an extra view's host **carries its own id and uses that** rather than the one passed through it, which is what stops a dispatch aimed at one output landing on another.

The implicit view keeps the tree and source already in place, so a shell with one view per engine — every Wayland configuration — takes the same path it always did. Reading the view id is gated on `struct_size`, so an engine older than that member reports every batch as the implicit view, which is what such an engine can produce.

## The test corrected me

I asserted the first view's node would be left behind but unreachable. It is removed outright — `Update()` prunes detached children. The test now asserts what actually happens, which is the more destructive of the two.

It exists because this is invisible without two outputs attached, and it is the constraint that makes a tree per view the only workable shape.

## Verification, and its limit

Verified on **Wayland**, where the behavior must not change:

- integration suite **8 of 8**
- two bundles still publish two trees
- 29/29 unit suites, clang-tidy clean, clang-format last

**The DRM path that reaches the new branch was not run.** It needs a card with two outputs, which this environment does not have. The routing, the lazy source registration and the per-view dispatch id are therefore reviewed-but-unexercised end to end — worth a reviewer's attention on hardware before this is relied on.

### DIFF
--- a/shell/engine.cc
+++ b/shell/engine.cc
@@ -355,6 +355,14 @@ void Engine::Shutdown() {
       m_engine_state->semantics_source = nullptr;
     }
   }
+  // Additional views go with it, for the same reason and at the same point.
+  for (auto& [view_id, view] : m_extra_views) {
+    if (view->source != nullptr) {
+      ihs_semantics_source_unregister(view->source);
+      view->source = nullptr;
+    }
+  }
+  m_extra_views.clear();
 #endif
   // Deinitialize stops new frame work; Shutdown joins the platform, UI and
   // rasterizer threads. After this returns nothing in the engine touches the
@@ -1092,6 +1100,7 @@ void Engine::InstallSemanticsHost() {
   // together, because publishing one without the other is the defect.
   if (m_engine_state != nullptr) {
     m_engine_state->semantics_source = m_semantics_source;
+    m_engine_state->engine = this;
   }
 }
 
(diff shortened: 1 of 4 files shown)


jwinarske:
Verified on hardware after merge — the DRM two-output limitation noted in this PR is closed.

A Pi 4 driving HDMI-A-1 and DSI-1 from one engine over `drm-kms-egl`:

```
[FlutterView] view 0 is on output 'HDMI-A-1'
[FlutterView] AddView id=1 800x480 result=0
(0) view 1 publishes as 'mcp-drive-test.1'

trees: ui://semantics/mcp-drive-test/tree      (status.view=0)
       ui://semantics/mcp-drive-test.1/tree    (status.view=1)

ui_tap Save on view 1 -> view 1 events 0->1, view 0 unchanged
```

Two Flutter views in one engine, each publishing its own tree, and an action reaching only the view it names. Both displays confirmed rendering.

This needed a fixture change to exercise: it called `runApp`, which renders into the implicit view alone, so the extra view produced no semantics at all. That change and the pointer-tap half are in #486.

PR #493, opened 2026-08-25
fix: stop DrmSession waking 5x/s on an idle system

### LINKED ISSUE #304: DrmSession: 200 ms timed poll wakes 5x/s on an idle system
`DrmSession`'s dispatch loop wakes 5 times a second on a fully idle system for
no reason other than to re-check a flag. It is the same anti-pattern #284 removed
from the input seats, on a different thread.

`shell/backend/drm_kms_egl/drm_session.cc:185`:

```c++
const int r = poll(pfds, nfds, kPollTimeoutMs);   // kPollTimeoutMs = 200 (line 37)
```

The loop polls the libseat fd and the hotplug fd, both of which are readable
exactly when there is something to do. The 200 ms timeout exists only so the
`while (!stop_.load(std::memory_order_acquire))` condition gets re-evaluated —
so the thread pays 5 wakeups/s forever to notice a flag that changes once, at
shutdown.

## Measured

On an RK3588 (NanoPC-T6) running `v2.0` with the DRM KMS EGL backend, 60 s idle,
counting `voluntary_ctxt_switches` per thread:

| Thread | wakeups / 60 s | rate |
|---|---|---|
| `li-drm-dispatch` (fixed by #284) | 0 | 0/s |
| the `DrmSession` poll thread | 301 | ~5/s |

5/s is exactly the 200 ms timeout. The number is identical on both sides of
#284, which is expected — that change never touched this loop — and it is what
now dominates the process's idle wakeups on the DRM path.

## Suggested fix

The same shape as #284: give the loop an eventfd waker, poll with no timeout,
and have `Stop()` (and any other cross-thread flag) write to the waker so the
poll returns. `shell/input/wake_event_fd.{h,cc}` already exists and is
self-contained; `DrmSeat::DispatchLoop` is the worked example, including the
degraded fallback to a timed poll when `eventfd()` is unavailable.

The producer set here looks smaller than the seats' — `stop_` is the obvious one;
worth checking whether anything else flips state that this loop is expected to
notice without an fd becoming readable.

## Notes

Out of scope for #261 by design; filed so it is not lost. Worth confirming
whether the software backend has an equivalent session loop before closing.

jwinarske:
Fixes #304.

## Problem

`DrmSession::DispatchLoop` polled with a 200 ms cap:

```cpp
const int r = poll(pfds, nfds, kPollTimeoutMs);   // kPollTimeoutMs = 200
```

Both fds it watches — the libseat fd and the hotplug fd — are readable exactly when there is something to dispatch. The timeout served only to re-evaluate `while (!stop_.load(...))`, a flag that changes once, at teardown. The thread paid five wakeups a second, forever, on a fully idle system.

## Change

Give the loop a `WakeEventFd` and block indefinitely (`timeout_ms = -1`). The destructor stores `stop_`, then wakes:

```cpp
DrmSession::~DrmSession() {
  stop_.store(true, std::memory_order_release);
  waker_.Wake();
  if (thread_.joinable()) { thread_.join(); }
```

Store-then-wake closes the lost-wake window — the eventfd stays readable across the dispatcher's check/block boundary, so a `Wake()` landing between its flag check and its `poll()` is still seen. That reasoning is documented on `WakeEventFd` itself.

This reuses `shell/input/wake_event_fd.h` rather than introducing a mechanism: `drm_seat` and `software_seat` already drive their loops this way, which is what #284 established.

**Degraded mode is preserved.** When `eventfd()` is unavailable `fd()` returns -1 and the loop falls back to the old 200 ms cap, so teardown is still observed promptly. `kPollTimeoutMs` stays, now purely as that fallback.

### Why the poll set moved to explicit indices

Not cosmetic. The old loop hard-coded `pfds[0]`/`pfds[1]` and tested `if (nfds == 2)` to decide whether the hotplug slot was populated. Hotplug can be dropped mid-loop on a bad fd (`hotplug_.reset(); hp_fd = -1;`), so once a third entry exists the waker slides from slot 2 into slot 1 — and `nfds == 2` would then read the waker as the hotplug fd and call `hotplug_->dispatch()` on a reset monitor.

The set is now rebuilt each iteration with `seat_idx` / `hp_idx` / `wake_idx`, matching `software_seat`.

## Verified on a Raspberry Pi 5

Cross-built with `emb` and run on the device against the real DRM-KMS-EGL backend.

Pi 5 Model B Rev 1.0, kernel `6.18.33+rpt-rpi-2712`, PiOS trixie arm64, `vc4`.

```
emb -v cross . --target rpi5-trixie --build --backend drm-kms-egl \
    --app test/integration/scroll_bench -m release \
    --deploy joel@raspberrypi.local --run
```

Builds clean for aarch64, and the shell reports its own provenance on startup:

```
SHEL: jw/drm-session-waker @ 5f684f3
[backend] resolved '<unset>' -> 'drm-kms-egl'
[drm] auto-selected first card with a connected display: /dev/dri/card0
[DrmDisplay] libseat session active
[DrmDisplay] opened /dev/dri/card0 via libseat (fd=7)
[DrmBackend] connector=35 crtc=94 mode=1280x1440@60Hz
[DrmBackend] driver='vc4' compositor=planes modeset=atomic explicit-sync=yes
```

So this is a real libseat session on a real DRM master — the exact fd the loop services, not a stand-in.

### The loop really does block indefinitely

The `DrmSession` thread is unnamed, so it is identified by its syscall rather than by `comm` (`li-drm-dispatch` is `DrmSeat`, set in `shell/input/drm_seat.cc`). On aarch64:

```
DrmSession thread: syscall=73 (ppoll)  arg2 (timespec*) = 0x0
```

A NULL `timespec` is an indefinite block. Unpatched, glibc would pass a pointer to a 200 ms `timespec` here, so this is kernel-side confirmation the change is in effect rather than an inference from the source.

### Idle wakeups

Context switches for that thread across 20 s of idle:

```
  start:          vol=19  nonvol=0
  after 20s idle: vol=19  nonvol=0
  DELTA:          voluntary=0  nonvoluntary=0
```

Zero. Unpatched this would be ~100 voluntary (5/s x 20 s). The thread blocks once and stays blocked.

### Teardown

`SIGTERM -> exit in 231 ms`, with no hang in `thread_.join()` — the specific runtime risk of moving to an indefinite poll.

## Also checked

Slot assignment was proven separately in a standalone harness mirroring the construction verbatim, since it is the part most likely to go wrong:

| hotplug | waker | nfds | seat | hp | wake |
| --- | --- | --- | --- | --- | --- |
| present | present | 3 | 0 | 1 | 2 |
| absent/dropped | present | 2 | 0 | -1 | 1 |
| present | absent (degraded) | 2 | 0 | 1 | -1 |
| absent | absent | 1 | 0 | -1 | -1 |

Each case asserts every populated slot holds the fd it claims, that no two indices alias, and that `nfds <= 3`. A dispatch harness on the same hardware serviced 150/150 events sub-millisecond across all three fd configurations, including after dropping hotplug mid-loop.

`clang-format --dry-run -Werror` clean; no lines over 80 columns.

## Not covered

Only the DRM-KMS-EGL backend was built and run. A VT switch (libseat pause/resume) and a live connector plug/unplug were not exercised — both drive `DrmSession` through paths this run left idle.

### DIFF
--- a/shell/backend/drm_kms_egl/drm_session.cc
+++ b/shell/backend/drm_kms_egl/drm_session.cc
@@ -120,6 +120,10 @@ DrmSession::DrmSession(drm::session::Seat seat) : seat_(std::move(seat)) {
 
 DrmSession::~DrmSession() {
   stop_.store(true, std::memory_order_release);
+  // Store the flag, then wake. That order closes the lost-wake window: the
+  // eventfd stays readable across the dispatcher's check/block boundary, so a
+  // Wake() landing between its stop_ check and its poll() is still seen.
+  waker_.Wake();
   if (thread_.joinable()) {
     thread_.join();
   }
@@ -172,17 +176,33 @@ void DrmSession::DispatchLoop() {
     return;
   }
   int hp_fd = hotplug_ ? hotplug_->fd() : -1;
+  const int wake_fd = waker_.fd();
   constexpr short kErr = POLLERR | POLLHUP | POLLNVAL;
 
+  // Block indefinitely. Both the seat and the hotplug fd are readable exactly
+  // when there is something to dispatch, so a timeout would only serve to
+  // re-check stop_ -- and stop_ changes once, at teardown, where the
+  // destructor wakes us instead. Degraded mode (eventfd() unavailable, fd()
+  // is -1) keeps the old cap so teardown is still observed promptly.
+  const int timeout_ms = wake_fd >= 0 ? -1 : kPollTimeoutMs;
+
   while (!stop_.load(std::memory_order_acquire)) {
-    pollfd pfds[2];
-    pfds[0] = {seat_fd, POLLIN, 0};
-    nfds_t nfds = 1;
+    // Indices rather than fixed slots: hotplug can be dropped mid-loop on a
+    // bad fd, and the waker is absent in degraded mode, so the set is rebuilt
+    // each iteration and the slots move.
+    pollfd pfds[3];
+    nfds_t nfds = 0;
+    const int seat_idx = static_cast<int>(nfds);
+    pfds[nfds++] = {seat_fd, POLLIN, 0};
+    const int hp_idx = hp_fd >= 0 ? static_cast<int>(nfds) : -1;
     if (hp_fd >= 0) {
-      pfds[1] = {hp_fd, POLLIN, 0};
-      nfds = 2;
+      pfds[nfds++] = {hp_fd, POLLIN, 0};
     }
-    const int r = poll(pfds, nfds, kPollTimeoutMs);
+    const int wake_idx = wake_fd >= 0 ? static_cast<int>(nfds) : -1;
+    if (wake_fd >= 0) {
+      pfds[nfds++] = {wake_fd, POLLIN, 0};
+    }
+    const int r = poll(pfds, nfds, timeout_ms);
     if (r < 0) {
       if (errno == EINTR) {
         continue;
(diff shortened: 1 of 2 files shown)


jwinarske:
## Measured on a Pi 5

Raspberry Pi 5 Model B Rev 1.0, kernel `6.18.33+rpt-rpi-2712`, aarch64.

The loop was reproduced verbatim — same poll-set construction, same index bookkeeping, same store-then-wake teardown order — and run in both variants back to back on the real scheduler, with both watched fds quiet for the whole run.

```
idle for 15s, both watched fds quiet

  old (200 ms cap)    polls=75   wakeups/s=5.00   dispatches=0   teardown=15.385 ms
  new (eventfd, -1)   polls=1    wakeups/s=0.07   dispatches=0   teardown=0.040 ms
```

The baseline reproduces the report exactly: **5.00 wakeups/s** on an idle system, 75 polls in 15 s, zero of which had anything to dispatch. With the waker the thread blocks once and stays blocked — the single poll is the initial block, so idle wakeups are gone rather than reduced.

Teardown improves as a side effect worth noting: the old loop cannot notice `stop_` until its current poll times out, so it waits out the remainder of the 200 ms window (15.4 ms in this run; up to 200 ms, ~100 ms on average). Store-then-wake returns in **0.040 ms**.

## Dispatch still works

Idle numbers are only half the question — the loop still has to service events. Signalling both fds and checking every event lands:

```
  seat events (hotplug present)      50/50 serviced, max latency 0.020 ms
  hotplug events                     50/50 serviced, max latency 0.004 ms
  hotplug dropped mid-loop; waker slot moves 2 -> 1
  seat events (hotplug dropped)      50/50 serviced, max latency 0.004 ms
  teardown 0.051 ms
ALL DISPATCH CHECKS PASSED
```

150/150 serviced across all three configurations, sub-millisecond throughout.

The third case is the one that matters for review: it drops the hotplug fd mid-loop, exactly as the error path does with `hotplug_.reset(); hp_fd = -1;`. The waker then moves from slot 2 to slot 1, and the seat keeps dispatching. Under the old `if (nfds == 2)` test that shift would have made the loop read the waker as the hotplug fd and call `hotplug_->dispatch()` on a reset monitor — which is why the poll set moved to explicit indices rather than that being tidy-up.

## What this does not cover

The Pi is a deployment target here, not a build host — no `cmake`, no `git`, and the `~/ivi-homescreen*` directories are deployed bundles (binary + `lib` + `data`) rather than source. So this exercises the loop's mechanism on the real kernel and scheduler, not the patched shell built in tree.

Still outstanding, and it needs a cross-build and a deployed bundle:

- that the change compiles in tree
- integration with the real libseat and `drm-cxx` hotplug fds rather than stand-ins
- a full session lifecycle: VT switch pause/resume, connector plug/unplug, and clean shutdown with no hang in `thread_.join()`

PR #498, opened 2026-08-28
logging: make the per-thread ring depth configurable

jwinarske:
## What

The per-thread SPSC ring is 256 slots, fixed at compile time. Depth now comes from `IHS_LOG_RING_CAPACITY`, defaulting to that same 256 — a shipped configuration is byte-for-byte unchanged unless the variable is set.

## Why

256 slots is a reasonable steady-state figure and a bad one for a burst. A producer that outruns the drain worker sheds records, and what survives reads like a complete log rather than one with holes in it.

Measured on a bounded 30k-record burst from one thread — the shape a scene load or asset-streaming pass has:

| capacity | emitted | dropped | |
|---|---|---|---|
| 256 (default) | 30000 | 29744 | 99.1% lost |
| 1024 | 30000 | 28976 | 96.6% lost |
| 8192 | 30000 | 21808 | 72.7% lost |
| 32768 | 30000 | 0 | |
| 65536 | 30000 | 0 | |

The cost is not the missing lines, it is that absence stops meaning anything. Comparing an asset manifest against the load records in a verbose run of a Flutter application said **19 of 114 textures never loaded**, one of them next to the visual defect being chased. Every one of them had loaded. Two runs of the same binary disagreed about *which* were missing, and that disagreement was the only thing that gave it away.

The same run with the depth raised keeps all **1,065,438** records, drops none, and accounts for all 106 textures.

## How

- `IHS_LOG_RING_CAPACITY` is clamped to `[16, 65536]` and rounded up to a power of two. The ceiling is ~20 MiB per logging thread — enough for a diagnostic run, low enough that a fat-fingered value is not an allocation failure. Anything adjusted or unparseable is reported on stderr, since a ring silently one power of two off from the request is exactly what later gets read back out of a log as evidence.
- Slots move from an inline array to a heap allocation, so `kRingCapacity`/`kMask` become per-ring members. `RingSlot` is `alignas(64)`, so array `new` still returns cache-line aligned storage and the inline array's false-sharing property holds.
- `push()` trades two compile-time constants for two `const` members on the same read-only cache line, against a `clock_gettime` and a 240-byte `memcpy` it already pays.
- Resolved once per process, on first use. Rings are pooled and reused across threads, so a depth that could change would leave rings of assorted depths behind.
- Construction can now throw. Rings are only created through `RingRegistry::acquire_ring()`, which already reaches the caller through a throwing `new ThreadRing()`.

## Tests

`compat_smoke` derives its burst from the depth instead of hardcoding 128, and asserts the bounds, the power-of-two invariant, and that a legal request is honored exactly — a knob that is read but not applied would leave an operator believing they had raised a ceiling that was still 256. The compat matrix registers a second run with the variable set; that needs a separate process, because the depth resolves once.

Built and passing locally at C++17/20/23 × `ENABLE_DLT=ON`/`OFF` (8 legs, `-Wall -Wextra -Werror`), plus adversarial values: `1000` → 1024, `3` → 16, `999999999` → 65536, `131072` → 65536, and `-1`/`abc`/`0` → default, each with the stderr notice and a clean exit.

## Follow-up, not in this PR

Nothing reports drops on the reading side — `ihs_shared` counts them per ring but never emits the count, so a reader has no signal that what they are holding is incomplete. A one-line report piggy-backed on the next successful enqueue is cheap and would have made the investigation above unnecessary. Happy to send it separately.

### DIFF
--- a/.github/workflows/ihs-logging-compat-matrix.yml
+++ b/.github/workflows/ihs-logging-compat-matrix.yml
@@ -19,6 +19,7 @@ on:
       - 'cmake/cxx_compat.cmake'
       - 'cmake/logging_helpers.cmake'
       - 'shell/logging/**'
+      - 'shared/src/logging/**'
       - '.github/workflows/ihs-logging-compat-matrix.yml'
   push:
     branches:
@@ -28,6 +29,7 @@ on:
       - 'cmake/cxx_compat.cmake'
       - 'cmake/logging_helpers.cmake'
       - 'shell/logging/**'
+      - 'shared/src/logging/**'
       - '.github/workflows/ihs-logging-compat-matrix.yml'
   workflow_dispatch:
 
--- a/shared/src/logging/thread_ring.cpp
+++ b/shared/src/logging/thread_ring.cpp
@@ -3,8 +3,10 @@
 
 #include <algorithm>
 #include <cstdio>
+#include <cstdlib>
 #include <cstring>
 #include <ctime>
+#include <memory>
 
 namespace {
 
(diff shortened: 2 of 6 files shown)


jwinarske:
Two notes on CI.

**Added a second commit** (`ci: run the logging compat matrix when the logging sources change`). The compat matrix's path filter watches `shell/logging/**`, but the harness compiles the ring, drain worker, bridge, context cache and sinks from `shared/src/logging/` — the code it exists to test sits outside its trigger. This PR only runs those legs because it happens to touch the tests and README alongside the ring; the ring-depth assertions it adds would not have run on a later change to the ring itself. Added `shared/src/logging/**` to both the `pull_request` and `push` filters.

Not changed, but worth a look separately: the `push` trigger lists `v2.0` and `jw/dlt-rebase`, so the matrix never runs on `v3.0`.

**`crash-handler` is failing, and it is not this PR.** It dies before reaching the crash path:

```
homescreen: error while loading shared libraries: libihs_shared.so.1:
    cannot open shared object file: No such file or directory
error: expected crash exit 139 (SIGSEGV), got 127
```

The shell links `../shared/libihs_shared.so.1.0.0` from the build tree with `-rpath $ORIGIN/lib:$ORIGIN`, and the harness runs it from a directory where neither resolves. The same failure, with the same message, hit `jw/wlxml-sysroot-without-cmake-sysroot` on 2026-08-27; `integration-test` last passed on 08-26. Nothing here touches the library name, soname, or install layout.

All logging legs pass: C++17/20/23 × `ENABLE_DLT=ON`/`OFF`, plus the DLT load test under both ASan and TSan.

PR #501, opened 2026-08-31
[docs] Per-backend READMEs, updated to v3.0

kerberjg:
## Summary

Introduces comprehensive architecture and operations documentation for every backend in `backend`. Each README covers the backend's purpose, architecture diagram, module responsibilities, feature matrix, build instructions, and runtime configuration.

Additionally, the backends are documented to match the current state of `v3.0`, which is ahead of this branch. The upstream changes are noted so the doc and code converge when the branches are reconciled.

## Backends documented

| Backend | README | v3.0 sync notes |
|---|---|---|
| `drm_kms_egl` | `README.md` | Dispatch loop redesign — `WakeEventFd`-based indefinite blocking poll |
| `drm_kms_vulkan` | `README.md` | — |
| `headless_egl` | `README.md` | — |
| `headless_vulkan` | `README.md` | — |
| `hud` | `README.md` | — |
| `software` | `README.md` | `drm_dumb_sink.h` now explicitly includes `<array>` |
| `wayland_egl` | `README.md` | `egl.h` now explicitly includes `<array>` |
| `wayland_leased_drm` | `README.md` | — |
| `wayland_vulkan` | `README.md` | `wayland_vulkan.h` now explicitly includes `<functional>` |

## Upstream changes documented

Two upstream commits on `v3.0` are not yet on this branch. Both are reflected in the relevant READMEs so that when the code is reconciled, the docs are already accurate.

### `drm_kms_egl` — `5f684f3a` *"fix: stop DrmSession waking 5x/s on an idle system"*

The `DrmSession` dispatch loop no longer polls every 200 ms. It now:

- Blocks indefinitely on a `WakeEventFd` (reuses `wake_event_fd.h`).
- Writes the eventfd once in the destructor after setting `stop_`, closing the lost-wake window across the check/block boundary.
- Falls back to the 200 ms cap only when the eventfd is unavailable (`fd() == -1`).

### `software` / `wayland_egl` / `wayland_vulkan` — `24e01fe3` *"fix: include the standard headers these declarations need"*

Three headers now include the standard headers they actually use, rather than relying on transitive includes:

| Header | New include |
|---|---|
| `drm_dumb_sink.h` | `<array>` |
| `egl.h` | `<array>` |
| `wayland_vulkan.h` | `<functional>` |

On Yocto dunfell (riscv64 / armv7), older toolchains don't re-export `std::array` or `std::function` from `<atomic>`, `<vector>`, `<EGL/egl.h>`, or `<mutex>`, so the build fails with `implicit instantiation of undefined template '…'`. Each affected README carries a short **Toolchain compatibility** note recording the root cause so the explicit includes aren't simplified away in future cleanups.

### DIFF
--- a/shell/backend/README.md
+++ b/shell/backend/README.md
@@ -1,0 +1,301 @@
+# Backend abstraction
+
+The display-target layer of the embedder. A `Backend` owns everything that is
+specific to *how* one Flutter view reaches the screen — surface lifecycle, the
+`FlutterRendererConfig` / `FlutterCompositor` handed to the engine, and the
+per-target vsync source — behind a single abstract interface
+([`backend.h`](backend.h)). One process can compile in several concrete
+backends; a process-wide [`BackendRegistry`](backend_registry.h) holds
+the compiled-in set, and the `Backend::Create` factory builds the right
+concrete instance per view at runtime from the resolved config. This folder is
+the parent of the per-backend implementation directories, each of which has its
+own README.
+
+## Features
+
+| Capability | Status | Toggle / scope |
+|---|---|---|
+| Abstract `Backend` interface (surface, renderer/compositor config, vsync) | Built-in | Always |
+| Runtime backend registry (`BackendRegistry` singleton) | Built-in | Always |
+| Per-view factory (`Backend::Create`) | Built-in | Always |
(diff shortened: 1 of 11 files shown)


kerberjg:
Above feedback addressed

jwinarske (approved):

PR #503, opened 2026-09-01
Drop the Firebase plugin build options from the README

jwinarske:
The plugins were removed in ivi-homescreen-plugins#268, so
`BUILD_PLUGIN_CLOUD_FIRESTORE`, `BUILD_PLUGIN_FIREBASE_AUTH` and
`BUILD_PLUGIN_FIREBASE_STORAGE` no longer exist — setting any of them now
does nothing.

Also drops the "Firestore first party compatible" feature bullet.

Documentation only; nothing in this repo's build referenced the removed
plugins.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ Flutter Linux CPP Embedder
     * Flutter Pigeon CPP compatible
     * Plugins modeled after Window CPP
     * Plugins enabled/disabled via CMake
-    * Firestore first party compatible
 * Desktop Texture Registry
     * Camera first party compatible
     * Video Player first party compatible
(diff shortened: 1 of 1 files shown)


jwinarske:
These three are in `generate_module_docs.py`, which this PR doesn't touch — it only drops seven lines from the README.

Fixed them in #504 instead: context manager for the file, `"-" * len(key)` for the underline, and the unused `subprocess` import removed. Ran both versions over the same doxygen XML and diffed the generated `modules.rst` to confirm the output is unchanged.

#504 is stacked on this one, so merge this first.

PR #510, opened 2026-09-11
docs: state the CMake flag defaults the build actually resolves

aki1770-del:
Four rows in the CMake Build flags table on v3.0 disagree with what a configure actually resolves, and one flag the table documents is declared nowhere.

Two commands check the whole change:

```
$ cmake -B build -G Ninja
$ grep -E '^(ENABLE_AGL_SHELL_CLIENT|BUILD_EGL_ENABLE_MULTISAMPLE|ENABLE_SIMPLE_SHELL_CLIENT):' build/CMakeCache.txt

$ git log --all -S ENABLE_DRM_LEASE_CLIENT -- '*.cmake' CMakeLists.txt
```

The table says AGL defaults OFF, but the cache resolves it ON. It says multisample defaults ON, but the cache resolves OFF. Simple-shell is in the cache with no row at all.

The second command returns nothing. `ENABLE_DRM_LEASE_CLIENT` has never been an option in any revision — it entered as a README line in `9ceb43fa` and stayed one. The capability is real under `BUILD_BACKEND_WAYLAND_LEASED_DRM`, which this README already documents in the Backend Support table, so the row now points there rather than restating the spec.

Defaults are read from `CMakeCache.txt` rather than from the `option()` call, because `option()` alone is not sound in this tree: `cmake/options.cmake:106` declares `BUILD_HUD` ON and `:107-110` forces it OFF whenever `BUILD_COMPOSITOR` is OFF, which is the default.

Bounds: I verified the 22 rows declared in this repository. The 15 `BUILD_PLUGIN_*` rows are declared outside it and I did not check them.

Documentation only — no build file is touched and no default changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,11 +148,13 @@ See the [configuration docs](shell/configuration/README.md) for the complete opt
 
 `ENABLE_XDG_CLIENT` - Enable XDG Client. Defaults to ON
 
-`ENABLE_AGL_SHELL_CLIENT` - Enable AGL Client. Defaults to OFF
+`ENABLE_AGL_SHELL_CLIENT` - Enable AGL Client. Defaults to ON
 
 `ENABLE_IVI_SHELL_CLIENT` - Enable ivi-shell Client. Defaults to OFF
 
-`ENABLE_DRM_LEASE_CLIENT` - Enable drm lease Client. Defaults to OFF
+`ENABLE_SIMPLE_SHELL_CLIENT` - Enable RDK/Westeros simple_shell Client. Defaults to OFF
+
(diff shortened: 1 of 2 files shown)


jwinarske:
Thanks @aki1770-del for finding these. Replaced by #512.

- `ENABLE_AGL_SHELL_CLIENT` -> OFF (matches README)
- `BUILD_EGL_ENABLE_MULTISAMPLE` stays OFF; README fixed
- `ENABLE_DRM_LEASE_CLIENT` dropped from docs
- `ENABLE_SIMPLE_SHELL_CLIENT`, `BUILD_BACKEND_WAYLAND_LEASED_DRM` added to docs

v3.0 is still churning, so for anything like this please open an issue first; saves rework on both ends.

aki1770-del:
Thanks for #512 — changing the AGL default is the better fix, and you're right that multisample was the README's error, not the build's. You also fixed the VULKAN row; we checked 22 rows and read past it.

Understood on v3.0: issues first from here.

PR #555, opened 2026-09-12
Add a tessella-quad target for a trixie Raspberry Pi 5

jwinarske:
One manifest, `.emb/tessella-quad.emb.yaml`, building the tessella quad for a Pi 5 on PiOS
trixie. Board → this project → app.

    emb cross . --target rpi5-trixie-quad -w <workspace> \
      --build --backend drm-kms-vulkan --app <tessella_fluorite>/example/quad

**Base is `wip/530-gl-release`, not `main`** — `main` carries no `.emb/`, so the layering this
target depends on does not exist there. Retarget freely; the manifest is one new file and depends
on nothing else in the branch.

### Why it is shaped this way

It lives in this repo rather than beside the app it builds because emb configures the embedder
from the directory holding the manifest, not from the project the manifest extends. A manifest in
the app's own tree sends cmake looking for a `CMakeLists.txt` in `.emb/`.

It extends the **board** rather than this project's `rpi5-trixie` target, because a project
reference to the project you are in is rejected as a cycle. The other `.emb` files merge in at the
cross level anyway, so the base augments and backends arrive regardless; only the target-level
`dev_packages` had to be named again, and they are kept in step with `raspberry-pi.emb.yaml`'s
`proj_packages_trixie`.

### Three things the layering does not carry

- **libc++-18 / libc++abi-18.** Filament compiles with clang and `-stdlib=libc++`; the ARM GNU
  toolchain ships libstdc++ only, so the C++ runtime comes from the target rootfs. Without them
  the Filament augment fails on `fatal error: 'string' file not found`, which reads like a broken
  sysroot and is a missing standard library.
- **gcc for the host-native wayland-cxx scanner.** It builds for the *host*, where the clang that
  Filament needs fails to compile the scanner's own headers against the system libstdc++. The
  augment is re-declared only to name its compilers.
- **`BUILD_COMPOSITOR: 'ON'`.** It lives in this project's target, not its base layer. Without it
  the FlutterCompositor API compiles out, `register_factory('views/fluorite-view')` fails `-4`,
  and every pane dies on `MissingPluginException` for `flutter/platform_views`.

### Checks

- `drm-kms-vulkan` shell and a release/arm64 app bundle both build from this target
- ran 38 minutes on a Pi 5: 88,680 frames, 38.5 fps sustained, four platform views composited per
  frame, zero errors, clean SIGTERM shutdown


jwinarske:
Wrong home. This manifest builds the tessella quad; it belongs in tessella, not in the shell it happens to run inside. Moving it there.

PR #568, opened 2026-09-14
test: confirm a card is vkms before handing it to a harness

jwinarske:
`ihs_find_vkms_card` identified vkms by its connector signature alone: a card advertising `cardN-Virtual-M` was taken to be vkms, on the reasoning that real GPUs do not use that name. Hyper-V's synthetic display does. On an Azure runner both the vkms card and the VM's own display match the pattern, and a first-match search returns whichever enumerated first — it has been returning the right one by luck of ordering, not by rule.

## The fix

A new `ihs_is_vkms_card` confirms the match against `cardN/device`, which resolves to `/sys/devices/faux/vkms` since vkms moved to the faux bus, and also accepts the older platform-device spelling whose driver symlink read `vkms`.

This is deliberately **not** a match on the driver name. On my local rig card0 is vkms with `driver=faux_driver` — the name that once identified it now identifies the bus glue, and would break again at the next rename. The device path is named for the device rather than the bus, so it survives. The connector-name test stays as the cheap first filter.

Verified against both local cards:

| card | device | driver | verdict |
|---|---|---|---|
| card0 | `vkms` | `faux_driver` | confirmed |
| card1 | `0000:16:00.0` | `amdgpu` | rejected |

`ihs_find_vkms_card` resolves `/dev/dri/card0`; `ihs_find_vkms_card 2` correctly declines, since the default instance has one connected output and `Writeback-2` is excluded as a capture target.

## Consolidation

`present_census.sh` and `input_event_driven.sh` each carried a private copy of the old search, weaker than the shared one — no connector count, and `-e` where the shared helper uses `-d`. Both now source `lib/drm_card.sh`. No private copy remains in the tree; all five KMS harnesses go through the shared helper.

## CI impact

The consolidation is **not** the CI-visible part — neither of those two harnesses is referenced by any workflow. The jobs that run this code are `drm-kms.yml`'s `vkms-smoke` (`test/drm_kms_vkms.sh --check`, `ubuntu-24.04`) and `pios.yml`'s `osgi_multi_bundle.sh` on the Pi. Both already sourced the shared helper, so they pick up the tightening through `drm_card.sh`.

On that runner card0 is genuinely vkms via the faux bus, so it should continue to resolve. Worth watching on the first run regardless, since this is the one job the change can affect.

Where a card can no longer be identified, the harnesses now `die` rather than proceed on a guess. That is the better failure: a wrong card is interpolated into `IVI_SW_SINK="drm-dumb:${VKMS_CARD}"`, and `sink_factory.cc:89-91` falls back to `NoneSink` with only a warning — so a misidentified card produces a census of nothing that reports as a pass. `VKMS_CARD` remains the documented override.

## Also

Corrects a comment in the discovery loop claiming connector child nodes are skipped for having no `device/` dir. They have one — a symlink back to the card — and pass that test; they are rejected by the connector-name test on the line below. The filtering was right, the stated reason was not.

`shellcheck` and `shfmt` are not installed here and no repo config or workflow invokes either, so this was checked with `bash -n` and exercised against the live sysfs tree.


jwinarske:
Correction to the CI-impact section above: it claims `present_census.sh` and `input_event_driven.sh` are "not referenced by any workflow" and that the consolidation is therefore not the CI-visible part. **That is wrong, and backwards.**

Both harnesses *are* exercised in CI, by `scripts/vkms_harness_integration.sh` — which `drm-kms.yml`'s `vkms-smoke` job runs in its "Harness integration (census + input harness on vkms)" step:

- `scripts/vkms_harness_integration.sh:101` — `env CENSUS_SECS=12 KEEPUP_MIN=0.30 test/present_census.sh`
- `scripts/vkms_harness_integration.sh:109` — `env IDLE_CTXT_MAX=150 test/input_event_driven.sh`

I reached the wrong conclusion by grepping `.github/workflows/` for the harness filenames. They are invoked through a wrapper script, so the grep came back empty and I read that absence as proof rather than as a limit of the search.

What this changes: the consolidation is not inert on CI. Those two harnesses previously resolved a card by connector name alone, and on that runner they now go through the confirmed `ihs_find_vkms_card`. That is the intended fix — on an Azure runner the name-only search can match Hyper-V's synthetic display — but it means this PR has a second CI-visible surface beyond `vkms-smoke`'s `--check` pre-flight, and the census/input legs are where a card that can no longer be identified would now `die` instead of proceeding.

The rest of the PR body stands. `VKMS_CARD` remains the documented override, and on this runner card0 is genuinely vkms via the faux bus.

PR #579, opened 2026-09-15
drm_kms_vulkan: refuse Impeller instead of presenting nothing

jwinarske:
Impeller's Vulkan backend requires a window-system presentation stack on the instance the shell hands the engine. This backend is KMS-direct — it renders into a dma-buf it scans out itself, creates no `VkSurfaceKHR`, and so declares no WSI extensions. Passing `--engine-arg=--enable-impeller` anyway fails inside the engine:

```
[ERROR:impeller/renderer/backend/vulkan/capabilities_vk.cc(105)]
      Impeller validation: Could not find the surface extension.
[ERROR:impeller/renderer/backend/vulkan/context_vk.cc(187)]
      Impeller validation: Device has insufficient capabilities.
[ERROR:embedder/embedder_surface_vulkan_impeller.cc(76)] Failed to initialize Vulkan Context.
[ERROR:shell/common/platform_view.cc(76)] Failed to create platform view rendering surface
```

Those come from the **engine's** logger, not the shell's, so nothing in `[E] SHEL:` marks the run as broken. The shell then carries on — libseat session active, DRM master held, cursor up, input started — and presents **zero frames**. A dead panel and a clean exit is the worst pair to debug.

## Why this refuses rather than probes

The requirement is not one missing extension that could simply be added. I tested each way of satisfying it:

| attempt | result |
| --- | --- |
| nothing enabled (current) | refused at the surface check, `capabilities_vk.cc:105` |
| `VK_KHR_surface` | refusal *moves* to the WSI check, `capabilities_vk.cc:154` |
| `VK_EXT_headless_surface` | not accepted as a WSI extension |
| `VK_KHR_display` | loader takes the display; the backend's own `drmSetMaster` then fails `EPERM` |

`VK_KHR_surface` is available on these systems — the shell simply never declares it. But satisfying the WSI check would mean declaring a window system this backend does not have. So the gate is **unconditional**: there is no capability here worth probing, and the refusal is a property of the backend's design, not of the driver.

Silently substituting Skia was the alternative and is worse — it runs the renderer the operator just asked not to use. This follows the rule the explicit-sync option already applies, where a demand that cannot be met fails loudly rather than quietly taking the fallback.

```
[E] [FlutterView] drm-kms-vulkan cannot run Impeller: its Vulkan backend requires a
    window-system (WSI) surface stack, and this backend presents directly to KMS with
    no window system. Drop --enable-impeller to run Skia, or use a backend that
    presents through a surface (wayland-vulkan).
[F] [App] view 0 failed to initialize; aborting
```

## Header move

The switch parse already existed for `wayland_egl`, which gates partial repaint on the same answer. A Vulkan-only build doesn't compile `wayland_egl`, so `backend/wayland_egl/partial_repaint_gate.h` was unreachable there and its include sat behind `#if BUILD_BACKEND_WAYLAND_EGL`. Moved to `shell/engine_switches.h` — same pure function, same unit test, now reachable from any backend regardless of which others are compiled in. The test directory keeps its name; only its include changed.

## Validation

| invocation | result |
| --- | --- |
| `--engine-arg=--enable-impeller` | refuses, aborts, no DRM master held |
| `--engine-arg=--enable-impeller=false` | runs, 9 frames, 0 errors |
| no switch | runs, 9 frames, 0 errors |

The `=false` case confirms the parse honors the value rather than blanket-matching the flag.

Gates: host build clean, Vulkan-only cross build clean (which is what proves the header move), `clang-tidy` clean, `format.sh --check` clean, `gen_config_reference.py --check` clean, and the 8 `PartialRepaintGate` tests pass through the moved header.

Found while measuring the renderer gap behind #578.


jwinarske:
Closing this — it is the wrong fix, and the premise behind it is false.

Impeller **is** reachable on this backend. I got it to initialize and announce itself:

```
[IMPORTANT:embedder_surface_vulkan_impeller.cc(80)] Using the Impeller rendering backend (Vulkan).
```

What it took, walking the capability checks in `capabilities_vk.cc` at the engine revision we ship:

| stage | wants | note |
| --- | --- | --- |
| instance, `:105` | `VK_KHR_surface` | available here; the shell just never declared it |
| instance, `:154` | one of a fixed WSI list | `VK_KHR_portability_enumeration` counts and implies no window system |
| device, `:285` | `VK_KHR_swapchain` | available; never actually used on the compositor path |

None of these are functionally used — the embedder renders into an image we hand it and scans it out ourselves. They are declarations a capability check demands.

Past those, Impeller runs and hits a real bug **in this shell**:

```
[ERROR:gpu_surface_vulkan_impeller.cc(145)] Invalid VkImage given by the embedder.
```

`VulkanDrmBackend::GetNextImageCb` returns `image = 0` with the comment *"Unreachable on the compositor path"*. That is true under Skia and false under Impeller, whose `gpu_surface_vulkan_impeller` calls `AcquireImage` for the root surface even with a compositor set. A Skia-era assumption, not an upstream limitation.

So refusing would have hard-coded a policy against a renderer Flutter is standardizing on, papered over a stale assumption in our own code, and kept refusing after the real fix landed. The problem it addressed — a dead panel with the failure visible only in the engine's logger, not `[E] SHEL:` — is real but small; the engine does print four ERROR lines, and this only turned that into an abort.

Replacing with work that makes Impeller run: declare the extensions the capability check requires, and fix `GetNextImageCb` for the compositor path.

The `shell/engine_switches.h` header move from this branch is independently useful and will come back with that.

Issue #597, opened 2026-09-18
platform_views: the Vulkan dma-buf import offers a hardcoded LINEAR format list it never checks against the device


jwinarske:
**Retracted as filed.** I claimed the hardcoded LINEAR list was why nothing imported. It was not — #598 was, on its own. With the padding from #598 and nothing else changed, stock v3.0 offers LINEAR, the producer allocates LINEAR, and the import succeeds with zero failures over a 6 s run. I had the two defects stacked and attributed both to the first one I found.

What survives is narrower, and I have not measured it failing.

## The narrow version

`HostQueryCapabilities` offers `kDmabufImportFormats` — four packed RGB fourccs, all `DRM_FORMAT_MOD_LINEAR`, hardcoded, never checked against the device. `DmabufVulkanImporter::Import` then creates the image with `VK_IMAGE_USAGE_TRANSFER_SRC_BIT | VK_IMAGE_USAGE_SAMPLED_BIT`.

On this board the device does not advertise either bit for LINEAR:

| modifier | `drmFormatModifierTilingFeatures` | SAMPLED_IMAGE | BLIT_SRC |
| --- | --- | --- | --- |
| `DRM_FORMAT_MOD_LINEAR` | `0xc980` | no | no |
| `DRM_FORMAT_MOD_BROADCOM_UIF` | `0xdd83` | yes | yes |

(Pi 5, Mesa 25.0.7, V3D 7.1.7, `VK_FORMAT_B8G8R8A8_UNORM`, via `vkGetPhysicalDeviceFormatProperties2` + `VkDrmFormatModifierPropertiesListEXT`.)

So we create a sampled image on a modifier the driver does not advertise as sampleable. That is invalid usage whatever the driver tolerates, and this one tolerates it — the frames import and composite. A validation-layer run would flag it; I have not done one.

## Why I am not just fixing it

The obvious fix is to query the device and offer only what passes. I wrote that, and it works, but it is the wrong trade on its own: on this board it drops LINEAR and leaves only UIF, which means a CPU producer — a camera, a video tile, anything filling pixels on the host — has to tile every frame. That is a real cost imposed to satisfy a rule nothing is currently enforcing.

Worth deciding rather than assuming:

- run under validation layers and see whether this is actually flagged
- if it is, prefer the probe-passing modifiers but keep the tolerated ones rather than dropping them
[...]

Issue #598, opened 2026-09-18
platform_views: a producer has no way to know its dma-buf must be padded past vkGetImageMemoryRequirements


jwinarske:
A producer that allocates exactly what `vkGetImageMemoryRequirements` asked for is short, and nothing tells it so. The driver checks the dma-buf against a padded size it works out itself, not against `VkMemoryAllocateInfo::allocationSize`, and refuses the import.

This is the sole cause of a platform view importing nothing on this hardware. I first filed #597 alongside it; that was wrong — with the padding below and nothing else changed, stock v3.0 offers LINEAR, the producer allocates LINEAR, and the import succeeds.

## Measured

Pi 5, Mesa 25.0.7, V3D 7.1.7, `drm-kms-vulkan`. 1280x1440 `XR24` `DRM_FORMAT_MOD_LINEAR`, one plane, stride 5120.

| | bytes |
| --- | ---: |
| `vkGetImageMemoryRequirements().size` | 7372800 |
| `lseek(plane_fd[0], 0, SEEK_END)` | 7372800 |

Exactly equal, and the import is refused with `VK_ERROR_INVALID_EXTERNAL_HANDLE`. Padding the producer's allocation by extra rows, frame geometry unchanged:

| pad rows | import failures / 3 s |
| ---: | ---: |
| 0 | 178 |
| 64 | 0 |
| 256 | 0 |

v3dv returns this before touching the kernel — `strace -f -e trace=ioctl` counts five `PRIME_FD_TO_HANDLE` for the whole process (the backend's own scanout setup) while hundreds of frames fail. That count is the quickest way to tell a pre-kernel rejection from the kernel refusing the buffer, and worth knowing before someone spends an afternoon on it like I did.

Separately: `gbm_bo_create_with_modifiers` reports the tiled modifier for a buffer it allocated at exactly `width * height * 4`, i.e. it does not apply the padding the modifier implies. A producer that trusts gbm is short by construction, not by carelessness. That part is Mesa's to answer for; the two below are ours.

## Two gaps

**The contract says nothing.** `platform_view.h` documents `IhsFrame` as width, height, per-plane offset and stride, and a producer reasonably concludes the allocation follows from those. It does not. The producer surface should say that a producer allocates to the padded size the granted modifier implies, and that the padding is not derivable from the frame geometry.

[...]

PR #608, opened 2026-09-19
platform_views: correct why the SAMPLED note defers

jwinarske:
Comment only. Corrects the deferral rationale in the note from #607.

It said gbm on that board cannot allocate UIF at all, so LINEAR was the only modifier a producer could make and filtering the offered set was not an option. Probing the allocator directly says otherwise -- it is UIF *with SCANOUT* that is refused, and the import path never asks for scanout, because the frame is sampled through Vulkan rather than handed to a plane.

Same board, renderD128, 1280x1440 XR24, `gbm_bo_create_with_modifiers2`:

| modifier | usage | result |
| --- | --- | --- |
| UIF | RENDERING | OK |
| UIF | none | OK |
| UIF | RENDERING\|SCANOUT | failed |
| LINEAR | RENDERING | OK |
| LINEAR | RENDERING\|SCANOUT | OK |

Probing with SCANOUT is the natural thing to try for a display buffer, and it fails, which is how "UIF unavailable" reads off one run.

Also run end to end in that shape earlier: with the capability query filtered to modifiers that pass the importability probe, the board offered UIF for all four packed RGB fourccs, the producer allocated it at usage 0x4, and 20 s at 60 submits/s imported with zero failures.

So the filter route works. Still deferred -- on cost now rather than impossibility:

- a UIF buffer comes back at the linear stride, so the producer over-allocates exactly as #598 describes
- a CPU producer pays the tiling on every `gbm_bo_map`/unmap, cheap for a small tile and not free at full screen

The `vkCmdCopyImage` route becomes the fallback rather than the plan.

Numbers are on #597.


jwinarske:
Follow-up measurement that inverts the conclusion here, on a fact neither of us had when this landed.

This PR left the note saying the filter route works and `vkCmdCopyImage` is the fallback. The filter does work -- your end-to-end run stands. But it does not close #597, because `choose_format` ends in `return req->formats[0]`: on no match the producer's first choice wins regardless of what capabilities offered. Filtering only redirects producers that pass **no** formats and take `caps->formats[0]`.

So after filtering, a producer that explicitly names LINEAR still gets it, still submits it, and the importer still creates a SAMPLED image on a modifier that does not advertise SAMPLED. The invalid usage stays reachable for exactly those producers.

And the redirect is not free for the ones it does move. Same board, renderD128, 1280x1440 ARGB8888, `GBM_BO_USE_RENDERING`, 60 iterations of map(WRITE)/fill/unmap:

| modifier | median | min | max |
| --- | ---: | ---: | ---: |
| LINEAR | 0.57 ms | 0.56 | 0.61 |
| UIF | 2.52 ms | 2.47 | 3.36 |

+1.95 ms per frame, 4.4x -- about 12% of a 60 Hz budget.

So the ordering is the other way round from what the note now says: the import-side fix is what closes the defect, for every producer, and the filter is an optimisation on top of it that lets the common path skip the copy. Alone, the filter costs ~2 ms a frame for CPU producers and still leaves the defect reachable.

Numbers and the `choose_format` detail are on #597. The in-tree note needs a third pass to match -- happy to put one up.

Issue #617, opened 2026-09-20
drm_kms_vulkan: the scanout backing store is sampled without SAMPLED usage


jwinarske:
The scanout backing store is created without `VK_IMAGE_USAGE_SAMPLED_BIT` and then sampled. Validation reports it two ways, 20 of each over a 20 s run:

```
vkCmdPipelineBarrier(): pImageMemoryBarriers[0].newLayout
(VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL) is not compatible with VkImage
usage flags VK_IMAGE_USAGE_TRANSFER_DST_BIT|VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT
```

```
vkUpdateDescriptorSets(): pDescriptorWrites[0].pImageInfo[0].imageView was
created with VK_IMAGE_USAGE_TRANSFER_DST_BIT|VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT,
but descriptorType is VK_DESCRIPTOR_TYPE_COMBINED_IMAGE_SAMPLER
```

Those usage flags are ours, twice, in `vulkan_backing_store.cc`:

```cpp
ic.usage = VK_IMAGE_USAGE_COLOR_ATTACHMENT_BIT | VK_IMAGE_USAGE_TRANSFER_DST_BIT;
```

`VUID-VkImageMemoryBarrier-oldLayout-01211` and `VUID-VkWriteDescriptorSet-descriptorType-00337`.

## Only with a platform-view layer

Same board, same binary, same layer, 20 s each:

| run | VUIDs |
| --- | ---: |
| device bundle, no platform view | **0** |
| `pv_bench`, one platform view at 60 submits/s | **40** |

So it is the composite path: with a layer to blend, something samples the backing store. Without one it is only ever a colour attachment and the missing bit never shows.

## Not the #597 family

Worth separating. On a V3D board the same harness reports four families, two of which are LINEAR-not-sampleable (`VkImageViewCreateInfo-usage-02274`, `vkCmdDraw-magFilter-04553`). Those are absent here -- Adreno's LINEAR does advertise `SAMPLED_IMAGE`, measured. These two are left over after that, and are a different defect: the image is missing a usage bit we never asked for, on a driver that would have granted it.

## The fix is not just adding the bit

[...]

PR #621, opened 2026-09-21
drm_kms_vulkan: give each layer its own KMS plane, behind a flag

jwinarske:
Option B for #617: stop blending layers that the display controller could fetch itself. Off by default (`IVI_DRMVK_PLANE_LAYERS`) — see the numbers, it is not a win everywhere.

## What it does

The backend committed **one** scene layer and blended the rest into it. Every platform view and every overlay backing store therefore cost a full-screen sample, plus — on a GPU whose scanout modifier is not sampleable — the copy added in #620. Both buffers are already scanout-capable by construction, so they can go on their own planes.

This reconciles one `drm::scene` layer per Flutter layer, keyed on the identity the engine already hands back (`store.get()`, `ICompositorSurface*`), with add / update-in-place / prune, zpos biased by the primary's own, and ARGB8888 so an overlay does not paint opaque black over the view below it.

**All-or-nothing**, and that is correctness rather than convenience: a layer blended into the bottom store sits *below* anything on an overlay plane, so mixing the two inverts the stack wherever a blended layer belongs above a placed one. A frame the allocator cannot place entirely goes back to the blend path, which is untouched.

## The part that was actually blocking it

`GetDmabuf` — the direct-scanout seam — was inside `#if IVI_HAVE_EGL`. On a Vulkan-only build the override did not exist, so the base class answered:

```cpp
[[nodiscard]] virtual DmabufState GetDmabuf(Dmabuf* /*out*/) const {
  return DmabufState::kNotScanoutCapable;
}
```

Every platform view reported "cannot scan out" on every frame, however many planes were free. Moving it out from behind that guard is most of what makes this work.

Its Vulkan half then needs something to offer. The importer consumes the submitted fds (#606), so the host now keeps a dup beside each import, in a cache with exactly that import's key and lifetime — retired with it, closed on teardown. Closing ours while a plane still scans is safe: the KMS framebuffer holds its own reference and the scene's pool dups what it keeps.

## Measured

Producer at 60 submits/s, 3 layers placed on both boards, no errors over a 25 s soak:

| | blend | planes | |
| --- | ---: | ---: | --- |
| V3D 7.1.7.0 | 61.51, 61.51 | **61.86, 61.77** | raster p50 16.13 -> **15.82 ms** |
| Adreno 640 | 51.5 | **48.8** | mean of five pairs |

**V3D gains, Adreno loses ~5%.** Scanning N buffers costs N times the display fetch, and on a pipe that is already the bottleneck — Adreno sits below vsync in this bench — that outweighs the GPU work saved. Caching the allocator's verdict against the layer set (a `TEST_ONLY` ioctl per frame otherwise) recovered part of it; the rest tracks the extra fetch.

So this is opt-in, not a default. The case it is aimed at is a tight frame budget where the GPU is the constraint, which is also where #620's copy hurts most: 1.75 ms at 1280x1440, 5.9 ms at 4K, per layer.

Still open: validation on a 240 Hz panel, which is the first configuration where the GPU saving and the fetch cost can be told apart properly. No display reachable here exceeds 60.3 Hz.

Gates: clang-tidy clean on the changed files, `scripts/format.sh`.


jwinarske:
Qualifying the Adreno number above, because I no longer trust it as a clean signal.

Chasing why that board sits below vsync at all, I profiled the present path there:

```
barrier=0.31ms  wait=11.48ms (max 18.40)  commit=0.80ms  total=12.60ms   fps=49.46
```

That run has **no producer and no platform view** -- one layer, no blend, no copy. It is *slower* than the loaded runs. Stripping the work out makes the frame rate worse, so nothing in the present path is the limiter there: the GPU contributes 0.3 ms of a 16.67 ms budget and the commit under 1 ms.

What is actually going on on that box: two vendor processes have been pegged at ~100% CPU since boot (9546 and 9553 minutes of CPU on a 6-day uptime), a third at 40%, load average 15 on 7 cores. The connector is a shared/virtualised DP behind a vendor display manager rather than a direct panel, and the flip wait swings 11.5-18.4 ms. For comparison, the same code on V3D waits 15.78 ms with a max of 15.79 -- 0.01 ms of variance.

So the V3D figures in the description stand (repeatable to about +/-0.05 fps). The Adreno pair -- 51.5 blend vs 48.8 planes -- was measured against that contention, and I would not conclude the ~5% gap is the display-fetch cost I attributed it to. It may be partly or wholly environmental. Re-measuring once that box is quiet is a prerequisite for reading anything into it.

Does not change what lands here: the flag is off by default either way, and the V3D result is the one that motivated it.

jwinarske:
Re-measured with the noisy process on that board stopped. The caveat above was wrong the wrong way round: the gap is bigger, not smaller.

Five pairs each, same binary, same bench:

| | mean fps | spread |
| --- | ---: | ---: |
| blend | 52.65 | 0.88 |
| planes | 48.51 | 2.13 |

Earlier numbers were blend 51.52 spread 2.65, planes 48.79 spread 2.88. Quieting the box moved blend up and tightened it 3x; planes did not move. So the ~8% is the plane path, not contention, and "may be partly or wholly environmental" was unfounded.

Nothing to change in what landed -- flag is off by default and V3D is what motivated it. But read the Adreno figure as a real cost on that display pipe: N planes, N times the fetch.

That board still tops out near 52 rather than 60 for unrelated reasons (a vendor process holds a core, and the connector is virtualised behind a display manager). Comparative numbers, not absolute.

